PR #1, opened 2016-05-31
Evgeniy/wsproxy dispatcher

evostrov:


README.md
```diff
@@ -1 +1,55 @@
-# perl-Mojo-WebSocketProxy
\ No newline at end of file
+# perl-Mojo-WebSocketProxy
+[![Build Status](https://travis-ci.org/binary-com/perl-Mojo-WebSocketProxy.svg?branch=master)](https://travis-ci.org/binary-com/perl-Mojo-WebSocketProxy) 
+
+Using this module you can forward websocket JSON-RPC 2.0 requests to RPC server.
```
jy-binary: can you beef up the README more? explain at a higher level about how it can be used

lib/Mojo/WebSocketProxy/Dispatcher/CallingEngine.pm
```diff
@@ -0,0 +1,255 @@
...
+    my $msg_type             = $req_storage->{msg_type};
+    my $rpc_response_handler = $req_storage->{response};
+
+    my $api_response = {
+        msg_type  => $msg_type,
+        $msg_type => $rpc_response,
+    };
+
+    # If RPC returned only status then wsapi will return no object
+    # TODO Should be removed after RPC's answers will be standardized
```
jy-binary: this TODO should be removed - it's an open-source module, either you support non-standard answers or you don't

lib/Mojo/WebSocketProxy/Dispatcher/Parser.pm
```diff
@@ -0,0 +1,100 @@
...
+        }
+        return $result;
+    }
+    return;
+}
+
+sub _failed_key_value {
+    my ($key, $value) = @_;
+
+    if ($key =~ /password/) {    # TODO review carefully
```
jy-binary: shouldnt see this in an open-source repo I think

README.md
evostrov: More detailed docs will be available on CPAN, from [this](https://github.com/binary-com/perl-Mojo-WebSocketProxy/pull/1/files#diff-cef47db8734c648526c6fe7587f28f9eR15) POD.
I wanted to paste a link to CPAN in README. 
Should I duplicate that doc in README?

jy-binary: no need to duplicate

lib/Mojo/WebSocketProxy/Dispatcher/CallingEngine.pm
evostrov: Just removed.
I think we can leave it.

lib/Mojo/WebSocketProxy/Dispatcher/Parser.pm
evostrov: Just changed it to skip_check_sanity regex

Makefile.PL
```diff
@@ -0,0 +1,41 @@
...
+WriteMakefile(
+    NAME             => 'Mojo::WebSocketProxy',
+    VERSION_FROM     => 'lib/Mojo/WebSocketProxy.pm',
+    AUTHOR           => 'Binary.com (binary@cpan.org)',
+    ABSTRACT         => 'Perl implementation of websocket-JSON-RPC 2.0 proxy',
+    PL_FILES         => {},
+    MIN_PERL_VERSION => 5.010,
+    BUILD_REQUIRES   => {
+        'Test::More'       => 0,
+        'Test::Mojo'       => 0,
```
basiliscos: this one comes with `mojo` already :)

README.md
basiliscos: @evostrov use

https://metacpan.org/pod/distribution/Pod-Markdown/bin/pod2markdown

to generate readme from pod

lib/Mojo/WebSocketProxy.pm
```diff
@@ -0,0 +1,252 @@
...
+            'web_socket_proxy' => {
+                actions => [
+                    [
+                        'json_key', 
+                        {
+                            instead_of_forward => sub {
+                                shift->call_rpc({  
+                                    args => $args,
+                                    method => $rpc_method, # it'll call 'http://rpc-host.com:8080/rpc_method'
+                                    rpc_response_cb => sub {...}
```
basiliscos: I would prefer shorter name, i.e. `cb` or `callback`

lib/Mojo/WebSocketProxy.pm
```diff
@@ -0,0 +1,252 @@
...
+    
+Global hooks which will run at the end of request handling.
+
+=head2 before_get_rpc_response (global)
+
+    before_get_rpc_response => [sub { my ($c, $req_storage) = @_; ... }, sub {...}]
+
+Global hooks which will run when asynchronous RPC call is answered.
+
+=head2 after_got_rpc_response (global)
```
basiliscos: @tfoertsch @evostrov 

Maybe generalize before/after hooks to `around`? We'll get twice less functions?

lib/Mojo/WebSocketProxy/CallingEngine.pm
```diff
@@ -0,0 +1,253 @@
...
+        return sub {
+            my $rpc_response = shift;
+            return $rpc_response_cb->($c, $req_storage->{args}, $rpc_response);
+        };
+    } else {
+        return sub {
+            my $rpc_response = shift;
+            if (ref($rpc_response) eq 'HASH' and exists $rpc_response->{error}) {
+                $error_handler->($c, $rpc_response, $req_storage) if defined $error_handler;
+                return error_api_response($c, $rpc_response, $req_storage);
```
basiliscos: Shouldn't it be 

```perl
return $error_handler->($c, $rpc_response, $req_storage) if defined $error_handler;
return error_api_response($c, $rpc_response, $req_storage);

```

or just 
```
my @args = ($c, $rpc_response, $req_storage);
return $error_handler ? $error_handler->(@args) : error_api_response(@args);

```

lib/Mojo/WebSocketProxy/CallingEngine.pm
```diff
@@ -0,0 +1,253 @@
...
+            } else {
+                $success_handler->($c, $rpc_response, $req_storage) if defined $success_handler;
+                store_response($c, $rpc_response);
+                return success_api_response($c, $rpc_response, $req_storage);
```
basiliscos: The same here, imho. If success_handler is defined it should be responsible for storing response, and creating output for the ws-client, i.e `return` should be used on the line 52

lib/Mojo/WebSocketProxy/CallingEngine.pm
```diff
@@ -0,0 +1,253 @@
...
+                $api_response = $c->wsp_error($msg_type, 'CallError', 'Sorry, an error occurred while processing your request.');
+                send_api_response($c, $req_storage, $api_response);
+                return;
+            }
+
+            $api_response = &$rpc_response_cb($res->result);
+
+            return unless $api_response;
+
+            if (length(JSON::to_json($api_response)) > ($max_response_size || 328000)) {
```
basiliscos: It's better to make the `328000` tunable configuration variable.

lib/Mojo/WebSocketProxy/CallingEngine.pm
```diff
@@ -0,0 +1,253 @@
...
+sub send_api_response {
+    my ($c, $req_storage, $api_response) = @_;
+
+    my $before_send_api_response_hook = delete($req_storage->{before_send_api_response}) || [];
+    my $after_sent_api_response_hook  = delete($req_storage->{after_sent_api_response})  || [];
+
+    $_->($c, $req_storage, $api_response) for @$before_send_api_response_hook;
+    $c->send({json => $api_response});
+    $_->($c, $req_storage) for @$after_sent_api_response_hook;
+    return;
```
basiliscos: Yeah, **around** looks natural here :)

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -0,0 +1,220 @@
...
+            }
+
+            $c->after_forward($result, $req_storage);
+        } elsif (!$result) {
+            $c->app->log->debug("unrecognised request: " . $c->dumper($args));
+            $result = $c->wsp_error('error', 'UnrecognisedRequest', 'Unrecognised request.');
+        }
+    };
+    if ($@) {
+        $c->app->log->info("$$ timeout for " . JSON::to_json($args));
```
basiliscos: You have used `warn` before, why now you have switched to `log->info` ?

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -0,0 +1,220 @@
...
+
+    # Enable permessage-deflate
+    $c->tx->with_compression;
+
+    my $config = Mojo::WebSocketProxy::Dispatcher::Config->new->{config};
+
+    Mojo::IOLoop->singleton->stream($c->tx->connection)->timeout($config->{stream_timeout}) if $config->{stream_timeout};
+    Mojo::IOLoop->singleton->max_connections($config->{max_connections}) if $config->{max_connections};
+
+    $config->{opened_connection}->($c) if $config->{opened_connection} and ref($config->{opened_connection}) eq 'CODE';
```
basiliscos: I think check `ref() eq 'CODE'` is too late here (every time, at runtime). That should be checked in plugin configuration time

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -0,0 +1,220 @@
...
+sub dispatch {
+    my ($c, $args) = @_;
+
+    my $log = $c->app->log;
+    $log->debug("websocket got json " . $c->dumper($args));
+
+    my ($action) =
+        sort { $a->{order} <=> $b->{order} }
+        grep { defined }
+        map  { Mojo::WebSocketProxy::Dispatcher::Config->new->{actions}->{$_} } keys %$args;
```
basiliscos: why do that at runtime? Why do not have actions in the config as array, and always pick the first?

lib/Mojo/WebSocketProxy/Dispatcher/Config.pm
```diff
@@ -0,0 +1,59 @@
+package Mojo::WebSocketProxy::Dispatcher::Config;
+
+use Mojo::Base -base;
+
+my $actions;
+my $config;
+
```
basiliscos: O, man! Don't use the global config, please. Store config and actions inside your Config object instance

lib/Mojo/WebSocketProxy/Dispatcher/Config.pm
```diff
@@ -0,0 +1,59 @@
+package Mojo::WebSocketProxy::Dispatcher::Config;
+
+use Mojo::Base -base;
```
basiliscos: use strict; use warnings ?

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -0,0 +1,220 @@
+package Mojo::WebSocketProxy::Dispatcher;
```
basiliscos: use strict; use warnings ?

lib/Mojo/WebSocketProxy.pm
```diff
@@ -0,0 +1,252 @@
+package Mojo::WebSocketProxy;
+
+use strict;
```
basiliscos: warnings? :)

lib/Mojolicious/Plugin/WebSocketProxy.pm
```diff
@@ -0,0 +1,278 @@
+package Mojolicious::Plugin::WebSocketProxy;
```
basiliscos: use strict; use warnings ?

lib/Mojolicious/Plugin/WebSocketProxy.pm
```diff
@@ -0,0 +1,278 @@
...
+
+use Mojo::Base 'Mojolicious::Plugin';
+use Mojo::WebSocketProxy::Dispatcher::Config;
+use Mojo::WebSocketProxy::Dispatcher;
+our $VERSION = "0.06";
+$VERSION = eval $VERSION;
```
basiliscos: What is the sense of that?

lib/Mojolicious/Plugin/WebSocketProxy.pm
```diff
@@ -0,0 +1,278 @@
+package Mojolicious::Plugin::WebSocketProxy;
+
+use Mojo::Base 'Mojolicious::Plugin';
+use Mojo::WebSocketProxy::Dispatcher::Config;
+use Mojo::WebSocketProxy::Dispatcher;
+our $VERSION = "0.06";
```
basiliscos: In the readme it is stated that vesion is 0.06 :) use pod2mkdown to keep them in synch

lib/Mojolicious/Plugin/WebSocketProxy.pm
```diff
@@ -0,0 +1,278 @@
...
+$VERSION = eval $VERSION;
+
+sub register {
+    my ($self, $app, $config) = @_;
+
+    die 'No actions found!' unless $config->{base_path};
```
basiliscos: Die message is incorrect

lib/Mojo/WebSocketProxy/Dispatcher/Parser.pm
```diff
@@ -0,0 +1,104 @@
...
+        $req_storage->{args} = {};
+        $result = $c->wsp_error('error', 'BadRequest', 'The application sent an invalid request.');
+    }
+
+    $result = _check_sanity($c, $req_storage) unless $result;
+
+    return $result;
+}
+
+sub _check_sanity {
```
basiliscos: @evostrov @tfoertsch 

What is the sense of checking sanity? 

Isn't it already checked at json decoding phase https://metacpan.org/source/SRI/Mojolicious-6.62/lib/Mojo/JSON.pm#L51 ?

I.e. if somebody would like transfer funny characters to rpc, like 

"☭", "⚒", "⚸", "⚳", "✪", "☢", "✉", "⚡", "☣"

why not?

lib/Mojo/WebSocketProxy.pm
evostrov: I think it's better to use long name to avoid get confused, because there could be many hooks and callback.

lib/Mojo/WebSocketProxy/CallingEngine.pm
evostrov: No, It's a hook to do additional actions when `success` or `error` response is returned. Then it does standard response handlers.

lib/Mojo/WebSocketProxy/CallingEngine.pm
evostrov: In your case it

evostrov: Your suggestion will lead to duplicate code if we want to do something additional action when success is returned and nothing change response. E.g. we have to write code to make standard response.

lib/Mojo/WebSocketProxy/CallingEngine.pm
evostrov: yes, thanks, nice catch

lib/Mojolicious/Plugin/WebSocketProxy.pm
evostrov: Just removed

lib/Mojo/WebSocketProxy/Dispatcher.pm
evostrov: > why do that at runtime?

We should do that at runtime anyway, regardless of using array or hash. Because we should do that for every message. In the normal way sort won't be run, because only one action will be find for the message.

>  Why do not have actions in the config as array, and always pick the first?

 Because getting value by hash key is faster than grep whole array every time, I think.

lib/Mojo/WebSocketProxy/Dispatcher/Config.pm
evostrov: Actually I won't like it :)
But what is wrong with it?
I'll use closure to save config.

basiliscos: If you create two config instances (e.g. for test or for different proxing), it will fail and would lead to hard-to-find bugs,

lib/Mojo/WebSocketProxy/Dispatcher/Parser.pm
evostrov: Maybe this comment helps https://github.com/binary-com/perl-Mojo-WebSocketProxy/pull/1/files#diff-e010619b1d5454359570e5ed3a0d624bR69

evostrov: As well as key validation https://github.com/binary-com/perl-Mojo-WebSocketProxy/pull/1/files#diff-e010619b1d5454359570e5ed3a0d624bR66

basiliscos: No. The question is - why we do sanity check for JSON at all? How JSON can be insane, and what errors might that introduce?

Now we validate json, the it has alpha-numeric keys and alpha-numeric-values with some subset of unicode symbols. The question is  - what is the sense of that validation at all?

lib/Mojo/WebSocketProxy/Dispatcher/Config.pm
evostrov: Fixed, please take a look

basiliscos: It still has shared `my $actions;` and `my $config`, and they are used in `init`

lib/Mojo/WebSocketProxy/Dispatcher/Parser.pm
```diff
@@ -0,0 +1,102 @@
...
+    return;
+}
+
+sub _failed_key_value {
+    my ($key, $value, $skip_check_sanity) = @_;
+
+    if ($skip_check_sanity && $key =~ /$skip_check_sanity/) {
+        return;
+    } elsif (
+        $key !~ /^[A-Za-z0-9_-]{1,50}$/
```
shuwnyuantee: @evostrov  @basiliscos  If you mean why we have sanity check at all here, it was already there when I started editing it. Somebody think is safer to only whitelist these char set.

Reason why L74 is added is, the check in L66 doesn't allow Japanese characters. So I added this exception.

If you think the while sanity check doesn't make sense, I am not against it. But check with @tfoertsch @raunakkathuria  @kavehmz

.travis.yml
```diff
@@ -0,0 +1,27 @@
+language: perl
+perl:
+    - "5.24"
```
tfoertsch: according to <https://travis-ci.org/binary-com/perl-Mojo-WebSocketProxy/jobs/135872261>, travis does not know about perl 5.24. This test is running with the system perl 5.14.

.travis.yml
```diff
@@ -0,0 +1,27 @@
...
+    - "5.22"
+    - "5.20"
+    - "5.18"
+    - "5.16"
+    - "5.14"
```
tfoertsch: you claim 5.10 compatibility below but don't test it.

tfoertsch: Do we need 5.10 compatibility? IMHO 5.10 is really old.

Makefile.PL
```diff
@@ -0,0 +1,40 @@
...
+    MIN_PERL_VERSION => 5.010,
+    BUILD_REQUIRES   => {
+        'Test::More'       => 0,
+        'Test::MockModule' => 0,
+        'Test::MockObject' => 0,
+        'JSON'             => 0,
+    },
+    PREREQ_PM => {
+        'Test::Simple' => 0.44,
+        'Mojolicious'  => 0,
```
tfoertsch: If I compare this list with the `before_install` section in `.travis.yml`, I get the impression that this here is not complete.

.travis.yml
evostrov: fixed

.travis.yml
evostrov: fixed

Makefile.PL
evostrov: fixed

PR #3, opened 2016-06-09
Minor fixes

evostrov:


lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -29,9 +34,7 @@ around 'send' => sub {
 };
 
 sub ok {
-    my $c      = shift;
-    my $source = 1;       # check http origin here
-    $c->stash(source => $source);
+    my $c = shift;
```
basiliscos: Please, remove this subroutine at all, if it is no needed any longer

PR #11, opened 2017-04-07
Add Future-support

basiliscos:


lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -66,40 +72,68 @@ sub on_message {
...
+                sub {
+                    my $next =
+                        $req_storage->{instead_of_forward}
+                        ? sub { $req_storage->{instead_of_forward}->($c, $req_storage) }
+                        : sub { $c->forward($req_storage) };
+                    Future->done($next->());
+                }
+                )->else(
+                sub {
+                    $result = shift;
```
chylli: `$result` not used ?

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -66,40 +72,68 @@ sub on_message {
...
+        )->then(
+        sub {
+            my $action = shift;
 
-    $c->send({json => $result}, $req_storage) if $result;
+            %$req_storage = (%$req_storage, %$action);
+            $req_storage->{method} = $req_storage->{name};
 
-    $c->_run_hooks($config->{after_dispatch} || []);
+            my $f = $c->before_forward($req_storage)->then(
```
chylli: is `$c->before_forward` a `Future` ?

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -66,40 +72,68 @@ sub on_message {
...
-                    ? $req_storage->{instead_of_forward}->($c, $req_storage)
-                    : $c->forward($req_storage);
-            }
-
-            $c->after_forward($result, $req_storage);
-        } elsif (!$result) {
-            $c->app->log->debug("unrecognised request: " . $c->dumper($args));
-            $result = $c->wsp_error('error', 'UnrecognisedRequest', 'Unrecognised request.');
+
+    my $result = Mojo::WebSocketProxy::Parser::parse_req($c, $req_storage);
```
chylli: Why not also wrap this line as `Future` ?

lib/Mojo/WebSocketProxy/Dispatcher.pm
basiliscos: It is the same `$result` which is defined in line #78. Yeah, that's silly, but that's original logic, and I was asked to keep backward compatibility.

lib/Mojo/WebSocketProxy/Dispatcher.pm
basiliscos: Yes, it  (and may others) are wrapped in future in `_run_hooks` (#162). I.e. if it does not returns future, `_run_hooks` will wrap it to immediately succeed or failed Future.

lib/Mojo/WebSocketProxy/Dispatcher.pm
basiliscos: It is not assumed, that the `parse_req` should be **asynchronous**. That's why `parse_req` should not return the Future. The `$result` is wrapped into future here, a few lines below.

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -66,40 +72,68 @@ sub on_message {
...
+                )->else(
+                sub {
+                    $result = shift;
+                    Future->fail;
+                });
+        }
+        )->then(
+        sub {
+            $result = shift;
+            $c->after_forward($result, $req_storage)->then(
```
basiliscos: It's a little bit crazy logic in the original code:

https://metacpan.org/source/BINARY/Mojo-WebSocketProxy-0.04/lib/Mojo/WebSocketProxy/Dispatcher.pm#L90

The only difference that in original code is executed always, after `before_forward`, meanwhile here it is executed only on success. 

`after_forward` uses the same `_run_hooks`, which can expect `Future`, i.e. leaves the possibility to do something asynchronously.

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -66,40 +72,68 @@ sub on_message {
...
-    };
-    if ($@) {
-        warn("$$ timeout for " . JSON::to_json($args));
-    }
+        )->then(
+        sub {
+            my $action = shift;
 
-    $c->send({json => $result}, $req_storage) if $result;
+            %$req_storage = (%$req_storage, %$action);
```
tfoertsch: This is better written as `@{$req_storage}{keys %$action} = values %$action;`

jy-binary:
@basiliscos pls fix conflict

PR #12, opened 2017-04-26
Drop Data::UUID

basiliscos:


lib/Mojo/WebSocketProxy/CallingEngine.pm
```diff
@@ -100,6 +99,8 @@ sub error_api_response {
     return $api_response;
 }
 
+my $request_no = 0;
```
chylli: Suggest to use `$request_number` instead.

raunakkathuria: will restart affect this any way? just want to be double sure

basiliscos: on restart it will reset $request_no to = 0, but as pid ($$) will be changed too, we'll get needed short-term uniquness.

PR #16, opened 2017-07-10
Add more details in warnings

basiliscos:


lib/Mojo/WebSocketProxy/CallingEngine.pm
```diff
@@ -144,7 +144,12 @@ sub call_rpc {
 
             my $api_response;
             if (!$res) {
-                warn "WrongResponse [$msg_type]";
+                my $tx = $client->tx;
+                my $details = 'URL: ' . $tx->req->url;
```
raunakkathuria: shouldn't this check if `tx` object exists, like `my $details = $tx ? 'URL: ' . $tx->req->url : '';`

basiliscos: Thanks for paying attention to that!

According to the implementation of `MojoX::JSON::RPC::Client` we shouldn't: 

- `$callback` is invoked at  https://metacpan.org/source/HENRYYKT/MojoX-JSON-RPC-0.10/lib/MojoX/JSON/RPC/Client.pm#L47 after doing `_process_result`

- in `_process_result` it already assumes `$tx` is defined https://metacpan.org/source/HENRYYKT/MojoX-JSON-RPC-0.10/lib/MojoX/JSON/RPC/Client.pm#L104 . I.e. if it will be undefined, it will `die` in non-our code already.

jy-binary:
@basiliscos failing tests

basiliscos:
Tests fail in master too, because `MojoX/JSON/RPC/Client.pm` uses Mojo6 api (as in our private cpan), meanwhile public tests use latest Mojo7, and it fails

PR #19, opened 2017-08-16
More details on failure

basiliscos:


lib/Mojo/WebSocketProxy/CallingEngine.pm
```diff
@@ -144,7 +144,12 @@ sub call_rpc {
 
             my $api_response;
             if (!$res) {
-                warn "WrongResponse [$msg_type]";
+                my $tx = $client->tx;
```
raunakkathuria: better to check if transaction exists as well

basiliscos: I think I have found the root of confusion: it is because comment above:

```perl
# unconditionally stop any further processing if client is already disconnected
return unless $c->tx;
```

The `$c->tx` and `$client->tx` are different transactions. The last one is between `ws-api` and RPC, the fitst one is connection of clients ot our `ws-api` via websockets.

According to the implementation of `MojoX::JSON::RPC::Client` the cheeck is useless:  

- `$callback` is invoked at  https://metacpan.org/source/HENRYYKT/MojoX-JSON-RPC-0.10/lib/MojoX/JSON/RPC/Client.pm#L47 after doing `_process_result`

- in `_process_result` it already assumes `$tx` is defined https://metacpan.org/source/HENRYYKT/MojoX-JSON-RPC-0.10/lib/MojoX/JSON/RPC/Client.pm#L104 . I.e. if it will be undefined, it will `die` in non-our code already.

In other words, `$client->tx` is outgoing transaction, and it cannot be `undef`. In case of RPC unavailability there will be error, but without `code`

PR #24, opened 2017-11-07
Preserve `req_id` after SanityCheckFailure

leonerd:


lib/Mojo/WebSocketProxy/Parser.pm
```diff
@@ -48,11 +48,6 @@ sub _check_sanity {
 
     if (@failed) {
         my $result = $c->wsp_error('sanity_check', 'SanityCheckFailed', 'Parameters sanity check failed.');
-        if (    $result->{error}
-            and $result->{error}->{code} eq 'SanityCheckFailed')
-        {
-            $req_storage->{args} = {};
```
chylli: Did you confirm with the author who added these lines ? Maybe there were some special reasons that we added these lines.

binarytom: It's from the initial commit and the author is no longer here. I think the idea was to clear out bad data so we don't send it back to the user, but this doesn't add much in the way of safety or security.

Since we were able to decode the original JSON data, we know that encoding it will be valid, so in protocol terms there's no strong reason for stripping out the data - plus including it makes things easier to debug.

PR #29, opened 2018-02-20
Support multiple backends

leonerd:
- [x] Support multiple backends
- [ ] Support different types of backend
- [ ] Define a backend using `Job::Async`


binarytom:
Merging as-is, we'll follow up with the different backend types on a new PR.

PR #30, opened 2018-03-05
Tom/backend job async

binarytom:
Support for Job::Async as a backend.

Also clears up some of the Future logic, since it was somewhat convoluted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,6 @@ matrix:
     fast_finish: true
 perl:
    - '5.24'
-   - '5.22'
-   - '5.20'
-   - '5.18'
-   - '5.16'
    - '5.14'
 before_install:
    - git config --global user.name "TravisCI"
--- a/lib/Mojo/WebSocketProxy/Backend.pm
+++ b/lib/Mojo/WebSocketProxy/Backend.pm
@@ -5,13 +5,9 @@ use warnings;
 
 no indirect;
 
-use Mojo::Base -base;
-
 use Mojo::WebSocketProxy::Backend::JSONRPC;
 
 ## VERSION
-
-has 'url';
 
 our %CLASSES = (
     jsonrpc   => 'Mojo::WebSocketProxy::Backend::JSONRPC',
@@ -25,9 +21,108 @@ sub class_for_type {
 
 sub new {
     my ($class, %args) = @_;
+    return bless \%args, $class unless $class eq __PACKAGE__;
+
     my $type = delete $args{type} or die 'need a backend type';
     my $backend_class = $class->class_for_type($type) or die 'unknown backend type ' . $type;
     return $backend_class->new(%args);
+}
+
+=head2 METHODS - For backend classes
+
+These will be inherited by backend implementations and can be used
+for some common actions when processing requests and responses.
+
+=cut
+
+sub make_call_params {
+    my ($self, $c, $req_storage) = @_;
+
+    my $args         = $req_storage->{args};
+    my $stash_params = $req_storage->{stash_params};
+
+    my $call_params = $req_storage->{call_params};
+    $call_params->{args} = $args;
+
+    if (defined $stash_params) {
+        $call_params->{$_} = $c->stash($_) for @$stash_params;
+    }
+
+    return $call_params;
+}
+
+sub get_rpc_response_cb {
+    my ($self, $c, $req_storage) = @_;
+
+    my $success_handler = delete $req_storage->{success};
+    my $error_handler   = delete $req_storage->{error};
+
+    if (my $rpc_response_cb = delete $req_storage->{rpc_response_cb}) {
+        return sub {
+            my $rpc_response = shift;
+            return $rpc_response_cb->($c, $rpc_response, $req_storage);
+        };
+    } else {
+        return sub {
+            my $rpc_response = shift;
+            if (ref($rpc_response) eq 'HASH' and exists $rpc_response->{error}) {
+                $error_handler->($c, $rpc_response, $req_storage) if defined $error_handler;
+                return error_api_response($c, $rpc_response, $req_storage);
+            } else {
+                $success_handler->($c, $rpc_response, $req_storage) if defined $success_handler;
+                store_response($c, $rpc_response);
+                return success_api_response($c, $rpc_response, $req_storage);
+            }
+            return;
+        };
+    }
+    return;
+}
+
+sub store_response {
+    my ($c, $rpc_response) = @_;
+
+    if (ref($rpc_response) eq 'HASH' && $rpc_response->{stash}) {
+        $c->stash(%{delete $rpc_response->{stash}});
+    }
+    return;
+}
+
+sub success_api_response {
+    my ($c, $rpc_response, $req_storage) = @_;
+
+    my $msg_type             = $req_storage->{msg_type};
+    my $rpc_response_handler = $req_storage->{response};
+
+    my $api_response = {
+        msg_type  => $msg_type,
+        $msg_type => $rpc_response,
+    };
+
+    if (ref($rpc_response) eq 'HASH' and keys %$rpc_response == 1 and exists $rpc_response->{status}) {
+        $api_response->{$msg_type} = $rpc_response->{status};
+    }
+
+    if ($rpc_response_handler) {
+        return $rpc_response_handler->($rpc_response, $api_response, $req_storage);
+    }
+
+    return $api_response;
+}
+
+sub error_api_response {
+    my ($c, $rpc_response, $req_storage) = @_;
+
+    my $msg_type             = $req_storage->{msg_type};
+    my $rpc_response_handler = $req_storage->{response};
+    my $api_response =
+        $c->wsp_error($msg_type, $rpc_response->{error}->{code}, $rpc_response->{error}->{message_to_client}, $rpc_response->{error}->{details});
+
+    if ($rpc_response_handler) {
+        return $rpc_response_handler->($rpc_response, $api_response, $req_storage);
+    }
+
+    return $api_response;
 }
 
 1;
--- a/lib/Mojo/WebSocketProxy/Backend.pm
+++ b/lib/Mojo/WebSocketProxy/Backend.pm
@@ -17,14 +17,6 @@ sub register_type {
     return;
 }
 
-sub class_for_type {
-    my ($class, $type) = @_;
-    my $target_class = $CLASSES{$type};
-    # extra () around require to avoid treating the function call as a class name
-    require(class_to_path($target_class)) unless $target_class->can('new');
-    return $target_class;
-}
-
 sub new {
     my ($class, %args) = @_;
     return bless \%args, $class;
@@ -32,7 +24,7 @@ sub new {
 
 sub backend_instance {
     my ($class, $type, %args) = @_;
-    my $backend_class = $class->class_for_type($type) or die 'unknown backend type ' . $type;
+    my $backend_class = $CLASSES{$type} or die 'unknown backend type ' . $type;
     return $backend_class->new(%args);
 }
 
--- a/lib/Mojo/WebSocketProxy/Backend.pm
+++ b/lib/Mojo/WebSocketProxy/Backend.pm
@@ -1,5 +1,4 @@
 package Mojo::WebSocketProxy::Backend;
-# TODO(leonerd): Split out JSON-RPC-specific parts into a subclass
 
 use strict;
 use warnings;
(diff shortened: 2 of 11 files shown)


lib/Mojo/WebSocketProxy/Backend.pm
```diff
@@ -1,20 +1,45 @@
...
+no indirect;
 
-use MojoX::JSON::RPC::Client;
-use Guard;
+use Mojo::Util qw(class_to_path);
 
 ## VERSION
 
-has 'url';
+our %CLASSES = (
```
leonerd: How would you feel about something like `Module::PluginFinder` here, rather than a statically configured list in one place?

Failing that, maybe have each of the specific modules register themselves into this base,

    package Mojo::WebSocketProxy::Backend::JSONRPC;
    use parent qw( Mojo::WebSocketProxy::Backend );
    __PACKAGE__->register_type( 'jsonrpc' );

That way, this common base doesn't need to know the names of specific types.

lib/Mojo/WebSocketProxy/Backend.pm
```diff
@@ -1,20 +1,45 @@
...
+    jsonrpc   => 'Mojo::WebSocketProxy::Backend::JSONRPC',
+    job_async => 'Mojo::WebSocketProxy::Backend::JobAsync',
+);
+
+sub class_for_type {
+    my ($class, $type) = @_;
+    my $target_class = $CLASSES{$type};
+    # extra () around require to avoid treating the function call as a class name
+    require(class_to_path($target_class)) unless $target_class->can('new');
```
leonerd: Either solution suggested above then avoids this problem, as `Module::PluginFinder` would find the packages, or the subclass-registers-itself just needs the containing program to `use` all the appropriate modules which then register themselves as a load-time side-effect.

lib/Mojo/WebSocketProxy/Backend.pm
binarytom: The registration approach would be sensible, yes - I know some people are dubious about autoloading everything available and just want to restrict to a subset, but that `->register_type` option is much neater than the current version.

lib/Mojo/WebSocketProxy/Backend.pm
```diff
@@ -1,20 +1,45 @@
...
+    return $target_class;
+}
+
+sub new {
+    my ($class, %args) = @_;
+    return bless \%args, $class unless $class eq __PACKAGE__;
```
leonerd: Subtle. base class `->new` to find the subclass which then calls that which probably `SUPER::` dispatches back here again. Probably wants splitting into two methods here; either

    sub _really_new { return bless ... }
    sub new { do type magic }

or maybe

    sub new { return bless ... }
    sub new_for_type { my ($class, $type, %args) = @_; ... }

Would a positional first argument `$type` be appropriate, or is the caller's structure firmly dependent on `type` being a regular `%args` key?

binarytom: Normally I'd see this as a case of bad class hierarchy - a manufactured unit should not inherit the entire factory, that's just asking for a machine uprising - but a separate `->backend_instance` method might be a good option here, we wouldn't expect external users to be messing around with this class anyway.

That'd mean we have `->backend_instance($type => %args)` and a regular `->new` method, with less chance of confusion anywhere in the callchain.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -0,0 +1,101 @@
...
+    # instance can be provided here.
+    $self->{client} //= do {
+        $jobman //= do {
+            # We don't hold a ref to this, since that might introduce unfortunate cycles
+            $loop //= do {
+                require IO::Async::Loop::Mojo;
+                IO::Async::Loop::Mojo->new;
+            };
+            $loop->add(
+                my $jobman = Job::Async->new
```
leonerd: This doesn't want to be `my`

lib/Mojo/WebSocketProxy/Config.pm
```diff
@@ -14,9 +14,9 @@ sub init {
     $self->{actions}  = {};
     $self->{backends} = {};
 
-    die 'Wrong parameter' if $in_config->{opened_connection} && ref($in_config->{opened_connection}) ne 'CODE';
-    die 'Wrong parameter' if $in_config->{finish_connection} && ref($in_config->{finish_connection}) ne 'CODE';
-    die 'Wrong parameter' if $in_config->{skip_check_sanity} && ref($in_config->{skip_check_sanity}) ne 'Regexp';
+    die 'Expected a coderef for opened_connection' if $in_config->{opened_connection} && ref($in_config->{opened_connection}) ne 'CODE';
+    die 'Expected a coderef for finish_connection' if $in_config->{finish_connection} && ref($in_config->{finish_connection}) ne 'CODE';
+    die 'Expected a regex for skip_check_sanity' if $in_config->{skip_check_sanity} && ref($in_config->{skip_check_sanity}) ne 'Regexp';
```
leonerd: General comment: I wonder if the general pattern of `ref $foo eq 'CODE'` ought to be changed in favour of `Ref::Util` - those tests are all efficient XS-or-custom-op code, so a bit more efficient than the native perl solution. Hardly an issue unique to this code, nor particularly important in this mere setup routine, but a general pattern we might want to consider for the codebase as a whole. Perhaps worthy of a new Trello card?

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -83,79 +82,66 @@ sub on_message {
     my $req_storage = {};
     $req_storage->{args} = $args;
 
-    my $result = Mojo::WebSocketProxy::Parser::parse_req($c, $req_storage);
+    # We still want to run any hooks even for invalid requests.
+    if(my $err = Mojo::WebSocketProxy::Parser::parse_req($c, $req_storage)) {
+        $c->send({json => $err}, $req_storage);
+        return $c->_run_hooks($config->{after_dispatch} || [])->retain;
+    }
+    my $result;
```
leonerd: This appears to be unused, and shadows a `my $result` later in the `->then` blocks

lib/Mojo/WebSocketProxy/Config.pm
binarytom: Yes, I like that idea. We were already pulling in XS deps (via Guard.pm, since removed) so I don't see any backcompat issues with adding that in now. :+1: for the Trello card to pick up other cases in our code too.

lib/Mojolicious/Plugin/WebSocketProxy.pm
```diff
@@ -58,14 +58,18 @@ sub register {
         die 'No actions found!';
     }
 
+    # For backwards compatibility, we always want to add a plain JSON::RPC backend
     $dispatcher_config->add_backend(default => Mojo::WebSocketProxy::Backend->new(
+        type => 'jsonrpc',
         url => delete $config->{url},
-    ));
+    )) unless exists $config->{backends}{default};
```
leonerd: This won't exist yet because the code adding them is the block just afterwards. So if the `unless` logic is wanted, it needs to be after the foreach loop currently just below.

t/25-job-async-backend.t
```diff
@@ -0,0 +1,135 @@
...
+
+package t::SampleWorker {
+    use parent qw(Job::Async::Worker);
+    use Future::Utils qw(repeat);
+
+    sub loop { shift->{loop} //= $loop }
+
+    sub start { }
+
+    sub trigger : method{
```
leonerd: Why does this need the `:method` attribute?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
binarytom: I think it needs to be, since this is the inner `do { ... }` block that will eventually be assigned via `$jobman //=`.

Was a bit dubious about this construct: I often use this pattern in my own code, but have grave doubts about the value of inflicting it on others. Pretty sure it would not be that hard to restructure to something more sensible.

lib/Mojo/WebSocketProxy/Backend.pm
leonerd: Yes - `PluginFinder` will `require` every possible module found on the filesystem, whereas a toplevel main program that `use`s just the ones it wants would be fine.

lib/Mojo/WebSocketProxy/Backend.pm
leonerd: Hmm.. something like

    my $backend_instance = Mojo::WebSocketProxy::Backend->class_for_type($type)->new(%args);

maybe?

lib/Mojolicious/Plugin/WebSocketProxy.pm
binarytom: Probably needed a better comment - the idea is that we'd allow people to provide an explicit `default => {...}` setting, allowing the removal of JSON::RPC if desired, but existing code should keep working.

Since it's using the config value, the data should be available - however! There is a clear bug here which I didn't catch earlier: the `exists` clause will autovivify `$config->{backends}` and mean the next `if()` clause which loops over those config values always runs. Will fix, thanks.

lib/Mojo/WebSocketProxy/Backend.pm
```diff
@@ -194,54 +138,6 @@ Mojo::WebSocketProxy::Backend
...
-=head2 get_rpc_response_cb
-
-=head1 SEE ALSO
-
-L<Mojolicious::Plugin::WebSocketProxy>,
-L<Mojo::WebSocketProxy>
-L<Mojo::WebSocketProxy::Dispatcher>,
-L<Mojo::WebSocketProxy::Config>
-L<Mojo::WebSocketProxy::Parser>
+See L<Mojo::WebSocketProxy::Backend> for the original JSON::RPC backend.
```
leonerd: ITYM `L<Mojo::WebSocketProxy::Backend::JSONRPC>`

t/25-job-async-backend.t
binarytom: It shouldn't, thanks - will remove. I tend to sprinkle those around the code at random just to pretend that attributes are sometimes useful... this particular code gains nothing from the addition.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
leonerd: Ah; in that case rather than `$jobman //= do { ...` maybe use

    unless($jobman) {
        ... code that side-effects assignment into $jobman
    }

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -0,0 +1,101 @@
...
+    )->then(sub {
+        my ($result) = @_;
+        $log->debugf('->submit completion: ', $result);
+
+        $_->($c, $req_storage) for @$before_get_rpc_response_hook;
+
+        # unconditionally stop any further processing if client is already disconnected
+        return Future->done unless $c and $c->tx;
+
+        my $mem_guard = guard {
```
leonerd: I am slightly dubious about the idea of using `Scope::Guard` in a Future chain. In this particular code it looks fine, but adding a second `->then` chain afterwards whose scope is supposed to contain the guard as well, would break it due to pad non-sharing of the `$mem_guard` lexical. I wonder if

    ->followed_by( sub { undef $req_storage; } )

would be better?
(Additionally I consider making `->finally` a name alias to `followed_by`)

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -0,0 +1,101 @@
...
+            undef $req_storage;
+        };
+
+        $_->($c, $req_storage, $result) for @$after_got_rpc_response_hook;
+
+#        if ($res->is_error) {
```
leonerd: Some code that appears to be commented out - is it dead? Can it be deleted?

lib/Mojo/WebSocketProxy/Config.pm
leonerd: https://trello.com/c/zYLQTFRi/6610-consider-refutil

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
binarytom: I think we should be able to tack an `->on_ready()` clause onto the end of the code, yes - there doesn't seem much benefit in `scope_guard` when we have timeout-controller Futures which are guaranteed to leave the `pending` state at some point.

t/25-job-async-backend.t
```diff
@@ -0,0 +1,135 @@
...
+            }
+            return Future->done;
+        } while => sub { 0 + @::PENDING_JOBS })->on_ready(sub {
+            delete $self->{active}
+        })
+    }
+
+    sub process {
+        my ($self, $job) = @_;
+        warn 'this one';
```
leonerd: Stray debugging `warn`?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
binarytom: Hopefully! I'd like to check how we're handling states such as RPC-error and transport-error, make sure that we always end up with something consistent and useful. basically "we had a response, but it was an error" should be a `->done` situation at the job level, so I wanted to leave this code intact to remind me to write test cases...

lib/Mojolicious/Plugin/WebSocketProxy.pm
leonerd: Ooooh right - that's the incoming config structure, not the one we're building up. I misread.

lib/Mojo/WebSocketProxy/Backend.pm
leonerd: Done 24f7ec3

lib/Mojo/WebSocketProxy/Backend.pm
leonerd: Done 99a543c

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
leonerd: Done e4027a1

lib/Mojo/WebSocketProxy/Dispatcher.pm
leonerd: Done 73bbca4

t/25-job-async-backend.t
leonerd: Done f6e7ed2

lib/Mojo/WebSocketProxy/Backend.pm
leonerd: Done 38aaf1a

.travis.yml
```diff
@@ -4,10 +4,6 @@ matrix:
     fast_finish: true
 perl:
    - '5.24'
-   - '5.22'
```
raunakkathuria: why all these versions are removed?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -0,0 +1,155 @@
...
+            $_->($c, $req_storage, $result) for @$after_got_rpc_response_hook;
+
+            $api_response = $rpc_response_cb->($result);
+        }
+        else {
+            my ($failure) = $f->failure;
+            $log->warnf("method %s failed: %s", $method, $failure);
+
+            $api_response = $c->wsp_error(
+                $msg_type, 'CallError', 'Sorry, an error occurred while processing your request.'
```
raunakkathuria: do we want to keep error message consistent? on top we send WrongResponse, here we send CallError

.travis.yml
binarytom: Since we don't use them internally and they're not officially supported, it seems wasteful to have travis trigger too many jobs here.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
binarytom: Yes, that might be better - will check to see why we had a different message. Thanks.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -0,0 +1,155 @@
...
+            );
+        }
+        return unless $api_response;
```
zakame: I suppose this is where a WrongResponse might be sent instead of returning, to be consistent with what `::JSONRPC` does?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
leonerd: That seems a good idea. The code currently is just a slightly edited and un-commented version of something that was in there beforehand, so I didn't really check if it was correct. I'll fix that up

leonerd: Done 861941b

PR #34, opened 2018-07-26
Dispatcher: normalize and decode WS_TEXT frames

zakame:
Mojo::Transaction::WebSockets emits `message` events with text frames preprocessed with Mojo::Util::decode, which renders any Unicode::Normalize work into no-ops.

    10> Mojo::Util::decode "UTF-8", "{\"ping\":\"\x{bf0}\"}"
    $res[4] = undef
    11> Mojo::Util::encode "UTF-8", "{\"ping\":\"\x{bf0}\"}"
    $res[5] = "{"ping":"௰"}"

Prevent this by subscribing to `text` events instead, which does not do any any preprocessing of the text frame.

Also catch possible exceptions from the JSON decoding, in a similar vein to the default `json` event.

### DIFF
--- a/lib/Mojo/WebSocketProxy/Dispatcher.pm
+++ b/lib/Mojo/WebSocketProxy/Dispatcher.pm
@@ -9,7 +9,7 @@ use Mojo::WebSocketProxy::Config;
 
 use Class::Method::Modifiers;
 
-use JSON::MaybeUTF8 qw(encode_json_utf8);
+use JSON::MaybeUTF8 qw(:v1);
 use Unicode::Normalize ();
 use Future::Mojo 0.004;    # ->new_timeout
 use Future::Utils qw(fmap);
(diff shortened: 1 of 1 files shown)


lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -64,7 +64,9 @@ sub open_connection {
         my ($c, $msg) = @_;
         # Incoming data will be JSON-formatted text, as a Unicode string.
         # We normalize the entire string before decoding.
-        my $args = eval { Mojo::JSON::decode_json(Unicode::Normalize::NFC($msg)) };
+        my $normalized_msg = Unicode::Normalize::NFC($msg);
+        my $args = eval { decode_json_utf8($normalized_msg) };
```
binarytom: I'd normally structure this as more of a try/catch block as in:

```
eval {
 my $args = decode...;
 on_message(...);
 1
} or do {
 log($@)
}
```

but I don't think we support (or expect much usage on) pre-5.14 Perl, and the decoding step failure modes aren't going to trigger the empty-`$@` issue so your current version is safe enough - the only remaining thing is that we're passing an `undef` value to `on_message()`, I'd be tempted to skip the call entirely.

However, if people are receiving invalid data here the current version does at least provide a mechanism to detect that and complain (e.g. drop the connection), so let's go with this as-is!

:1st_place_medal:

PR #36, opened 2018-08-10
Dispatcher: censor messages failing JSON decoding

zakame:
Fail the WebSocket connection immediately when receiving malformed input; log received messages on debug logging (if that level is active.)

Remove redundant test in t/05_basic.t as well, as this is covered now in t/14-bad-request.t.

https://tools.ietf.org/html/rfc6455#section-7.1.7

### DIFF
--- a/lib/Mojo/WebSocketProxy/Dispatcher.pm
+++ b/lib/Mojo/WebSocketProxy/Dispatcher.pm
@@ -65,9 +65,12 @@ sub open_connection {
         # Incoming data will be JSON-formatted text, as a Unicode string.
         # We normalize the entire string before decoding.
         my $normalized_msg = Unicode::Normalize::NFC($msg);
-        my $args = eval { decode_json_utf8($normalized_msg) };
-        $log->error("JSON decoding $normalized_msg failed: $@") if $@;
-        on_message($c, $args);
+        if (my $args = eval { decode_json_utf8($normalized_msg) }) {
+            on_message($c, $args);
+        } else {
+            $c->finish(1007 => 'Malformed JSON');
+            $log->debug(qq{JSON decoding failed for "$normalized_msg": $@});
+        }
     });
 
     $c->on(binary => sub {
--- a/t/05_basic.t
+++ b/t/05_basic.t
@@ -110,20 +110,6 @@ is_deeply $res,
     },
     'It should return error response if action does not exist';
 
-$t = $t->send_ok({json => 'not_json'})->message_ok;
-$res = decode_json_utf8($t->message->[1]);
-is_deeply $res,
-    {
-    'error' => {
-        'code'    => 'BadRequest',
-        'message' => 'The application sent an invalid request.'
-    },
-    'debug'    => 1,
-    'msg_type' => 'error'
-    },
-    'It should return error response if bad request';
-
-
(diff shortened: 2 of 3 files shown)


lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -66,7 +66,10 @@ sub open_connection {
         # We normalize the entire string before decoding.
         my $normalized_msg = Unicode::Normalize::NFC($msg);
         my $args = eval { decode_json_utf8($normalized_msg) };
-        $log->error("JSON decoding $normalized_msg failed: $@") if $@;
+        if ($@) {
```
binarytom: since this is a public module, I'd suggest:

```perl
if(my $args = eval { decode_json_utf8($normalized_msg) }) {
    on_message($c, $args);
} else {
    $log->debug("JSON decoding failed for: $normalized_msg");
    # maybe $c->emit(parse_error => $@, $normalized_msg);
    return;
};
```

zakame: Thanks @binarytom !  There doesn't seem to be a `parse_error` event currently; I was thinking of emitting a regular [`error`](https://metacpan.org/pod/release/SRI/Mojolicious-7.29/lib/Mojo/EventEmitter.pm#error) instead, would that work?

Another way I'm currently handling this is just to guard with [`is_level`](https://metacpan.org/pod/release/SRI/Mojolicious-7.29/lib/Mojo/Log.pm#is_level):

```perl
if (my $args = ...) {
    ...
} else {
    $log->debug(...);
    $log->error($@) if $log->is_level('debug');
}
```

What do think?  cc @raunakkathuria

zakame: OTOH, the `else` side of the above won't work (even with the `return;` as we do not emit a `BadRequest` response (which comes from `Dispatcher::on_message()/Parser::parse_req()`.

binarytom: I think better to have consistent logging - if it only appears at `debug` level, it should be a `debug` message.

If we receive a request which fails at the decoding step, I'd suggest just dropping the connection - we can send back a generic error first, "malformed JSON" for example.

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -66,7 +66,10 @@ sub open_connection {
...
+            $log->debug("JSON decoding failed for: $normalized_msg");
+            $log->error("JSON decoding error: $@") if $log->is_level('debug');
```
raunakkathuria: shouldn't this be also a debug message instead of checking debug level?

zakame: Originally the idea here is that we emit an error on the log only if the log level is debug, but I agree with @binarytom to just simplify this and emit to the debug log instead, foregoing any log level checks.

lib/Mojo/WebSocketProxy/Dispatcher.pm
zakame: Thanks, I've opted to follow the [spec](https://tools.ietf.org/html/rfc6455#section-7.1.7) on closing the connection here and logging to debug.

PR #44, opened 2019-01-07
Anthony wong/log wrong details type

Anthony-Wong:


lib/Mojolicious/Plugin/WebSocketProxy.pm
```diff
@@ -36,6 +38,10 @@ sub register {
                 message => $message
             };
             $error->{details} = $details if ref($details) eq 'HASH' && keys %$details;
+            
+            if ($details && ref($details) ne 'HASH') {
+                $log->debugf("Details in a websocket error must be a hash reference instead of a %s", ref($details));
+            }
```
zakame: Maybe `$details` could be tested only once like

```perl
if ($details && ref $details eq 'HASH' && keys %$details) {
    $log->debugf("Details in a websocket error must be a hash reference");
}
else {
    $error->{details} = $details;
}
```

Anthony-Wong: I assume you meant the if/else to be swapped, but that won't be entirely accurate though, as if details is missing, we shouldn't log any error. The only way I can think of to do this otherwise is with a nested if block (which may / may not make it messier). Thoughts?

turnstep: Think it's okay as is, FWIW.

PR #55, opened 2019-07-10
Mat/reset password to queue

mat-binary:


lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -66,7 +66,7 @@ sub new {
     # We'd like to provide some flexibility for people trying to integrate this into
     # other systems, so any combination of Job::Async::Client, Job::Async and/or IO::Async::Loop
     # instance can be provided here.
-    $self->{client} //= do {
+    unless $self->{client} {
```
zakame: Syntax error here, needs
```suggestion
    unless ($self->{client}) {
```

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -111,30 +111,30 @@ sub call_rpc {
...
     my $params = $self->make_call_params($c, $req_storage);
     $log->debugf("method %s has params = %s", $method, $params);
-
     $_->($c, $req_storage) for @$before_call_hook;
-
     $self->client->submit(
         name   => $req_storage->{name},
-        params => encode_json_utf8($params)
+        params => encode_json_utf8($params),
+        rpc_queue_client_tv => [Time::HiRes::gettimeofday],
```
raunakkathuria: what does `tv` stands for? Can we use descriptive name, please?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
nael-binary: IMHO this part doesnt need changing and should be kept as it is.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
nael-binary: i dont think we should be adding this extra parameter, to get stats that we actually already have.
we already have hooks in Websocket that will record for us the time spent in each call. In fact you can easily see now how long reset_password request takes to be served (attached screenshot)

Screenshot 2019-07-16 at 2 24 00 PM

so I think in here we should only have one `stats_inc` indicating a successful request to the queue.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -66,7 +66,7 @@ sub new {
     # We'd like to provide some flexibility for people trying to integrate this into
     # other systems, so any combination of Job::Async::Client, Job::Async and/or IO::Async::Loop
     # instance can be provided here.
-    $self->{client} //= do {
+    unless ($self->{client}) {
```
nael-binary: @mat-binary I dont see the need of changing this part

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-binary: @nael-binary I'm trying to cover what is missing in the existing metrics: `queue latency` (i.e. the interval between the moment a message is queued, until it is dequeued. I'm not sure if it is the correct term). How can we calculate it from the existing metrics? The closest (not exactly correct) answer is: `bom_websocket_api.v_3.rpc_call.timing` minus `bom_websocket_api.v_3.rpc_call.timing`.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-binary: Yes. It is a leftover of an experiment. I'll revert it.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
nael-binary: in this current way what its really measuring the network latency only as in (how long it took us from the moment we sent the request to the queue, until the worker picked it up. -not including the processing time.
I believe those are the two things that we are interested in, in terms of timing only:
- time spent in processing each job in worker (has to be added)
- time spent in order to serve the request to client. Hooks in WS are already in place for that.

why `bom_websocket_api.v_3.rpc_call.timing` is not correct? isnt it giving us the time spent on each RPC call, whether from normal RPC or job_async one?

mat-binary: @nael-binary I edited the last sentence of my message: 
```
The closest (not exactly correct) answer is: bom_websocket_api.v_3.rpc_call.timing minus bom_rpc.v_3.call.timing.
```
Anyway,I admit that keeping with the old metrics is more intuitive if we want to make before/after comparisons. We can measure queue performance by the existing metrics fairly (we cannot separate request queue latency from response latency, for example). 

It means that I can close two RPs. Hurray!

PR #57, opened 2019-07-29
Mat/reset password to queue

mat-fs:


lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -106,56 +88,79 @@ Implements the L<Mojo::WebSocketProxy::Backend/call_rpc> interface.
...
     my $msg_type = $req_storage->{msg_type} ||= $req_storage->{method};
-    $self->{client}->start->get;
 
-    $req_storage->{call_params} ||= {};
+    # We'd like to provide some flexibility for people trying to integrate this into
+    # other systems, so any combination of Job::Async::Client, Job::Async and/or IO::Async::Loop
+    # instance can be provided here.
+    $self->{client} //= do {
+        # We don't hold a ref to this, since that might introduce unfortunate cycles
+        my $loop //= do {
```
mat-fs: @nael-binary We are getting `perlcritic` warnings about this line. Isn't it better to be rewritten as `$self->{loop} = `?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -58,29 +58,11 @@ sub new {
     my ($class, %args) = @_;
     # Avoid holding these - we only want the Job::Async::Client instance, and everything else
     # should be attached to the loop (which sticks around longer than we expect to).
-    my $loop = delete $args{loop};
+    my $loop   = delete $args{loop};
     my $jobman = delete $args{jobman};
```
mat-fs: @nael-binary  These variables are not used anymore and just deleting from args.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
nael-binary: yes actually here its better to use `$self->{loop}`

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
nael-binary: this is needed here, not the variables themselves, but the deleting part for `$args` so maybe we can just delete without the need of assigning them to any variable

PR #61, opened 2019-08-18
mat/rpc_queue_connection_error_handling

mat-fs:


lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -13,6 +13,9 @@ use Job::Async;
 use JSON::MaybeUTF8 qw(encode_json_utf8 decode_json_utf8);
 use Log::Any qw($log);
 use MojoX::JSON::RPC::Client;
+use Syntax::Keyword::Try;
+
+my $PRC_QUEUE_TIMEOUT = $ENV{PRC_QUEUE_TIMEOUT} // 30;
```
michaelmueller-binary: is this meant to be $RPC_QUEUE_TIMEOUT ?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -137,19 +157,35 @@ sub call_rpc {
...
 
             if ($f->is_done) {
-                my $result = MojoX::JSON::RPC::Client::ReturnObject->new(rpc_response => decode_json_utf8($f->get));
-
-                $_->($c, $req_storage, $result) for @$after_got_rpc_response_hook;
-
-                $api_response = $rpc_response_cb->($result->result);
-                stats_inc("rpc_queue.client.jobs.success", {tags => ["rpc:" . $req_storage->{name}, 'clientID:' . $self->client->id]});
+                try {
+                    my $result = MojoX::JSON::RPC::Client::ReturnObject->new(rpc_response => decode_json_utf8($f->get));
```
michaelmueller-binary: in the non async version  `$result` is just set to empty if there was an issue with the call so will it thow an error now?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -137,19 +157,35 @@ sub call_rpc {
...
+
+                    $_->($c, $req_storage, $result) for @$after_got_rpc_response_hook;
```
michaelmueller-binary: if there was an issue with the call and no result was returned this is not called in the non async version.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-fs: Since this block is conditioned by `$->is_done`, there cannot be any issue here, right?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-fs: Again running if future is done. There is no failure here.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -54,6 +55,8 @@ C<< $jobman->client >> if not provided.
 
 =cut
 
+use constant RPC_QUEUE_TIMEOUT => $ENV{PRC_QUEUE_TIMEOUT} // 300;
```
raunakkathuria:
```suggestion
use constant QUEUE_TIMEOUT => $ENV{QUEUE_TIMEOUT} // 300;
```

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -120,11 +123,14 @@ sub call_rpc {
...
-        params => encode_json_utf8($params)
-        )->on_ready(
-        sub {
+
+    Future->wait_any(
+        $self->client->submit(
+            name   => $req_storage->{name},
+            params => encode_json_utf8($params)
+        ),
+        Future::Mojo->new_timer(RPC_QUEUE_TIMEOUT)->then(sub { Future->fail('Timeout') })
```
raunakkathuria: why not `$self->loop->timeout_future(after => QUEUE_TIMEOUT)->on_fail(sub { Future->fail('timeout') })

t/25-job-async-backend.t
```diff
@@ -12,13 +12,15 @@ use IO::Async::Loop::Mojo;
 
 our @PENDING_JOBS;
 
+$ENV{PRC_QUEUE_TIMEOUT} = 2;
```
raunakkathuria:
```suggestion
$ENV{QUEUE_TIMEOUT} = 2;
```

zakame: typo here (hence it seems better to just call it `QUEUE_TIMEOUT`)
```suggestion
$ENV{QUEUE_TIMEOUT} = 2;
```

t/25-job-async-backend.t
```diff
@@ -12,13 +12,15 @@ use IO::Async::Loop::Mojo;
...
+
 my $loop = IO::Async::Loop::Mojo->new;
 our $LAST_ID = 999;
 package t::SampleClient {
     use parent qw(Job::Async::Client);
 
     sub loop { shift->{loop} //= $loop }
-    sub start { Future->done }
+    sub start { shift->{startup_future} = Futute->done }
```
zakame:
```suggestion
    sub start { shift->{startup_future} = Future->done }
```

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-fs: Tests were failing with `$self->{loop} undefined` error. I refactored the code a little bit by adding a new `$self->loop` and it was fixed with your suggested change.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
michaelmueller-binary: ok

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
michaelmueller-binary: ok

t/25-job-async-backend.t
```diff
@@ -69,6 +71,7 @@ package t::FrontEnd {
          ( my $url2 = $url ) =~ s{/rpc/}{/rpc2/};
 
          my $client = t::SampleClient->new;
+         my $loop =
```
tom-binary: ?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -137,15 +153,24 @@ sub call_rpc {
...
+                catch {
+                    $log->warnf("Failed to process response of method %s: %s", $method, $_);
+                    stats_inc("rpc_queue.client.jobs.fail",
+                        {tags => ["rpc:" . $req_storage->{name}, 'clientID:' . $self->client->id, 'error:' . $_]});
+                    $api_response = $c->wsp_error($msg_type, 'WrongResponse', 'Sorry, an error occurred while processing your request.');
+                };
             } else {
-                my ($failure) = $f->failure;
-                $log->warnf("method %s failed: %s", $method, $failure);
+                my $failure = $f->is_failed ? $f->failure // '' : 'Rpc request was cancelled. Subscription is not ready.';
```
tom-binary:
```suggestion
                my $failure = $f->is_failed ? $f->failure // '' : 'RPC request was cancelled. Subscription is not ready.';
```

although I'd suggest splitting these out into separate done/failed/cancelled mappings, rather than putting it all under `->on_ready`

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -37,11 +38,11 @@ Returns a new instance. Required params:
 
 =over 4
 
-=item loop => IO::Async::Loop
+=item loop => IO::Async::Loop (deprecate)
```
tom-binary: why deprecated? shouldn't take any extra work to support these as parameters...

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -54,6 +55,8 @@ C<< $jobman->client >> if not provided.
 
 =cut
 
+use constant QUEUE_TIMEOUT => $ENV{QUEUE_TIMEOUT} // 300;
```
tom-binary: needs to be documented

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -13,6 +13,7 @@ use Job::Async;
 use JSON::MaybeUTF8 qw(encode_json_utf8 decode_json_utf8);
 use Log::Any qw($log);
 use MojoX::JSON::RPC::Client;
+use Syntax::Keyword::Try;
```
tom-binary: if we're adding new modules, they'd need to be listed in `cpanfile` as well

mat-fs: Added.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-fs: As I remember  @nael-binary removed these args because they harmful effect on queue client shutdown. I'll try to test service shutdown by restoring these args and push changes if there is not any issue. @nael-binary Would you please give details about the reason behind these args and the test procedure if want to restore them?

t/25-job-async-backend.t
mat-fs: Fixed.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-fs: Done.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-fs: @tom-binary I agree that splitting into done/failed/cancel is better, but doing so the lines of code shared among them should be duplicated which doesn't seem reasonable to me.

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
michaelmueller-binary: not changed?

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -91,64 +145,63 @@ sub call_rpc {
...
+            name   => $req_storage->{name},
+            params => encode_json_utf8($params)
+        ),
+        $self->loop->timeout_future(after => QUEUE_TIMEOUT)
+        )->on_ready(sub {
             my ($f) = @_;
             $log->debugf('->submit completion: ', $f->state);
 
-            $_->($c, $req_storage) for @$before_get_rpc_response_hook;
+            foreach my $hook (@$before_get_rpc_response_hooks) { $hook->($c, $req_storage) };
```
nael-binary: i think it was fine how it was. as in there is no need for this change

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
```diff
@@ -91,64 +145,63 @@ sub call_rpc {
...
-                my $result = MojoX::JSON::RPC::Client::ReturnObject->new(rpc_response => decode_json_utf8($f->get));
-
-                $_->($c, $req_storage, $result) for @$after_got_rpc_response_hook;
-
-                $api_response = $rpc_response_cb->($result->result);
-                stats_inc("rpc_queue.client.jobs.success", {tags => ["rpc:" . $req_storage->{name}, 'clientID:' . $self->client->id]});
+                try {
+                    $result = MojoX::JSON::RPC::Client::ReturnObject->new(rpc_response => decode_json_utf8($f->get));
+
+                    foreach my $hook (@$before_get_rpc_response_hooks) { $hook->($c, $req_storage, $result) };
```
nael-binary: ditto

lib/Mojo/WebSocketProxy/Backend/JobAsync.pm
mat-fs: @nael-binary There is nothing wrong with using `$_` here, but since there are a number of constructs around setting the default variable in different ways, I've accepted @michaelmueller-binary 's suggestion to use named variables for more readability.

PR #66, opened 2020-01-03
datadog/change_wrong_response_to_metrics

diego-binary:
- Change the log level for Wrong Response in JSONRPC to info instead of warning
- Send stats to DataDog for every Wrong Response in JSONRPC

### DIFF
--- a/Changes
+++ b/Changes
@@ -2,9 +2,8 @@ Revision history for Perl module Mojo::WebSocketProxy
 
 {{$NEXT}}
 
-          2019-09-20 01:44:27+00:00 UTC
-    - Modify log level for Wrong Responses from JSONRPC calls.
-    - Send wrong response statistics to DataDog.
+    - Remove warning line for RPC communication failure.
+    - Add documentation about RPC failure callback.
 0.13      2019-09-20 01:44:27+00:00 UTC
 
     - Change sanity error type
--- a/Changes
+++ b/Changes
@@ -2,6 +2,9 @@ Revision history for Perl module Mojo::WebSocketProxy
 
 {{$NEXT}}
 
+          2019-09-20 01:44:27+00:00 UTC
+    - Modify log level for Wrong Responses from JSONRPC calls.
+    - Send wrong response statistics to DataDog.
 0.13      2019-09-20 01:44:27+00:00 UTC
 
     - Change sanity error type
--- a/lib/Mojo/WebSocketProxy.pm
+++ b/lib/Mojo/WebSocketProxy.pm
@@ -130,7 +130,7 @@ Alternate backends are selected by using the C<backend> action option.
 
 =head2 rpc_failure_cb
 
-A sub routine reference to call if the remote call fails at a http level. 
+A subroutine reference to call when the RPC call fails at the HTTP level. 
 Called with C<< Mojolicious::Controller >> the rpc_response 
 and C<< $req_storage >>
 
@@ -147,7 +147,7 @@ A default rpc_failure_cb could be provided in the startup sub routine
                 url => 'http://rpc-host.com:8080/',
                 rpc_failure_cb => sub {
                     my ($c, $res, $req_storage) = @_;
-                    warn "Something went wrong an RPC call";
+                    warn "RPC call failed";
                     return undef;
                 }
 
--- a/lib/Mojo/WebSocketProxy.pm
+++ b/lib/Mojo/WebSocketProxy.pm
@@ -146,7 +146,7 @@ A default rpc_failure_cb could be provided in the startup sub routine
                 base_path => '/api',
                 url => 'http://rpc-host.com:8080/',
                 rpc_failure_cb => sub {
-                    my ($c, $res, $req_storage) = @_;
+                    my ($c, $res, $req_storage, $error) = @_;
                     warn "RPC call failed";
                     return undef;
                 }
@@ -163,7 +163,7 @@ Call specific sub routine could be specified in call_rpc arguments
         method         => 'ticks_history',
         rpc_failure_cb => sub {
             if ($worker) {
-                warn "Something went wront with this rpc call : ".$method;
+                warn "Something went wrong with this rpc call : " . $method;
                 $worker->unregister;
             }
         },
--- a/lib/Mojo/WebSocketProxy.pm
+++ b/lib/Mojo/WebSocketProxy.pm
@@ -128,6 +128,12 @@ supported.
 
 Alternate backends are selected by using the C<backend> action option.
 
+=head2 rpc_failure_cb
+
+A sub routine reference to call if the remote call fails at a http level. 
+Called with C<< Mojolicious::Controller >> the rpc_response 
+and C<< $req_storage >> 
+
 =head2 before_forward
 
     before_forward => [sub { my ($c, $req_storage) = @_; ... }, sub {...}]
--- a/lib/Mojo/WebSocketProxy.pm
+++ b/lib/Mojo/WebSocketProxy.pm
@@ -132,7 +132,43 @@ Alternate backends are selected by using the C<backend> action option.
 
 A sub routine reference to call if the remote call fails at a http level. 
 Called with C<< Mojolicious::Controller >> the rpc_response 
-and C<< $req_storage >> 
+and C<< $req_storage >>
+
+A default rpc_failure_cb could be provided in the startup sub routine
+
+    sub startup {
+        my $self = shift;
+        $self->plugin(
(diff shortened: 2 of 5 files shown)


lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -113,14 +119,16 @@ sub call_rpc {
...
                 my $api_response;
                 if (!$res) {
                     my $tx      = $client->tx;
                     my $details = 'URL: ' . $tx->req->url;
                     if (my $err = $tx->error) {
                         $details .= ', code: ' . ($err->{code} // 'n/a') . ', response: ' . $err->{message};
                     }
-                    warn "WrongResponse [$msg_type], details: $details";
+                    $log->info("WrongResponse [$msg_type], details: $details");
+                    stats_inc("rpc_queue.client.rpc.wrong_response", {tags => ["rpc:" . $req_storage->{name}, 'clientID:' . $self->client->id]});
```
raunakkathuria:
```suggestion
                    stats_inc("websocket.wrong_response", {tags => ["rpc:" . $req_storage->{name}, 'clientID:' . $self->client->id]});
```

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -120,7 +122,8 @@ sub call_rpc {
                     if (my $err = $tx->error) {
                         $details .= ', code: ' . ($err->{code} // 'n/a') . ', response: ' . $err->{message};
                     }
-                    warn "WrongResponse [$msg_type], details: $details";
+                    $log->info("WrongResponse [$msg_type], details: $details");
+                    stats_inc("websocket.wrong_response", {tags => ["rpc:" . $req_storage->{name}]});
```
zakame: We probably want a callback here so users of the code can supply their own metrics code.

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -11,6 +11,7 @@ no indirect;
 
 use curry;
 
+use DataDog::DogStatsd::Helper qw(stats_inc);
```
zakame: We can't use this directly here, otherwise we have to update the `cpanfile` dependency to require this module.  Instead, a callback should be supplied so that users of this module can introduce their own metrics.

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
zakame: Since we already have `$rpc_failure_cb` to call shortly, that can work here in user code:

```perl
$app->plugin(web_socket_proxy => {
    ...,
    rpc_failure_cb => sub {
        my ($c, $res, $req_storage) = @_;
        stats_inc(..., {tags => [ ..., $req_storage->{name} ]});
        return undef;
    },
    ...
});

File: Changes [hunk 4]
```diff
@@ -2,6 +2,9 @@ Revision history for Perl module Mojo::WebSocketProxy
 
 {{$NEXT}}
 
+          2019-09-20 01:44:27+00:00 UTC
```
zakame: Needs removing this timestamp (it should be supplied by the `{{$NEXT}}` line above automatically upon release.)

lib/Mojo/WebSocketProxy.pm
```diff
@@ -128,6 +128,48 @@ supported.
 
 Alternate backends are selected by using the C<backend> action option.
 
+=head2 rpc_failure_cb
+
+A sub routine reference to call if the remote call fails at a http level.
```
zakame:
```suggestion
A subroutine reference to call when the RPC call fails at the HTTP level. 
```

lib/Mojo/WebSocketProxy.pm
```diff
@@ -128,6 +128,48 @@ supported.
...
+        $self->plugin(
+            'web_socket_proxy' => {
+                actions => [
+                    ['json_key', {some_param => 'some_value'}]
+                ],
+                base_path => '/api',
+                url => 'http://rpc-host.com:8080/',
+                rpc_failure_cb => sub {
+                    my ($c, $res, $req_storage) = @_;
+                    warn "Something went wrong an RPC call";
```
zakame:
```suggestion
                    warn "RPC call failed";
```

By this point, `$res` should have something that could indicate the kind of error encounted (e.g. HTTP status code...) since we know further down that `$res->is_error`...

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -120,8 +122,8 @@ sub call_rpc {
                     if (my $err = $tx->error) {
                         $details .= ', code: ' . ($err->{code} // 'n/a') . ', response: ' . $err->{message};
                     }
-                    warn "WrongResponse [$msg_type], details: $details";
-                    $rpc_failure_cb->($c, $res, $req_storage ) if $rpc_failure_cb;
+                    $log->info("WrongResponse [$msg_type], details: $details");
+                    $rpc_failure_cb->($c, $res, $req_storage) if $rpc_failure_cb;
```
zakame: I wonder if supplying `$tx->error` (if present) would also be helpful here...

diego-binary: Hi @zakame, no sure if I am following correctly, but the `$details` string does already have some attributes from `$tx->error`, precisely `code` and `message`. When you wonder if it would be helpful do you mean to add some other attribute (or all of them) from `$tx->error` or if it will be better to just remove all `$tx->error` information at all?

Changes
diego-binary: ok,

lib/Mojo/WebSocketProxy.pm
diego-binary: 👍

Changes
tom-binary: The date shouldn't be included - it'll be autogenerated when this is released.

```suggestion
```

lib/Mojo/WebSocketProxy.pm
```diff
@@ -128,6 +128,48 @@ supported.
...
+
+Call specific sub routine could be specified in call_rpc arguments
+
+    $c->call_rpc({
+        args           => $args,
+        origin_args    => $req_storage->{origin_args},
+        method         => 'ticks_history',
+        rpc_failure_cb => sub {
+            if ($worker) {
+                warn "Something went wront with this rpc call : ".$method;
```
tom-binary:
```suggestion
                warn "Something went wrong with this rpc call: " . $method;
```

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -90,7 +92,7 @@ sub call_rpc {
     my $before_get_rpc_response_hook = delete($req_storage->{before_get_rpc_response}) || [];
     my $after_got_rpc_response_hook  = delete($req_storage->{after_got_rpc_response})  || [];
     my $before_call_hook             = delete($req_storage->{before_call})             || [];
-    my $rpc_failure_cb               =  delete($req_storage->{rpc_failure_cb}) || 0;
+    my $rpc_failure_cb = delete($req_storage->{rpc_failure_cb}) // undef;
```
tom-binary: The `// undef` here is a no-op: if the key isn't in the hash then `delete` will return `undef`.

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
tom-binary: @diego-binary you're not passing `$details` to the callback, though?

The point is that the action may vary depending on what exactly went wrong.

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -120,8 +122,8 @@ sub call_rpc {
                     if (my $err = $tx->error) {
                         $details .= ', code: ' . ($err->{code} // 'n/a') . ', response: ' . $err->{message};
                     }
-                    warn "WrongResponse [$msg_type], details: $details";
-                    $rpc_failure_cb->($c, $res, $req_storage ) if $rpc_failure_cb;
+                    $log->info("WrongResponse [$msg_type], details: $details");
```
tom-binary: The responsibility for logging should move to the failure function.

One option would be to provide the `$log` call as a default, so that the caller can override with something else.

PR #68, opened 2020-04-01
dont send if error is processed by failure cb

chylli-binary:


lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -119,14 +119,16 @@ sub call_rpc {
                     my $tx = $client->tx;
                     $req_storage->{req_url} = $tx->req->url;
                     my $err = $tx->error;
-                    $rpc_failure_cb->(
+                    my $cb_processed_totally;
+                    $cb_processed_totally = $rpc_failure_cb->(
```
billmarriott-binary, line 123: IIUC we are only using `rpc_failure_cb` in one other call. Why don't we just stop returning anything if a callback is defined? IMO this would make sense for all situations.

chylli-binary: Yes you are right. Will do it.

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -119,16 +120,15 @@ sub call_rpc {
...
+                    $rpc_failure_cb->(
                         $c, $res,
                         $req_storage,
                         {
                             code    => $err->{code},
                             message => $err->{message},
                             type    => 'WrongResponse',
                         }) if $rpc_failure_cb;
-                    return if $cb_processed_totally;
+                    return if $failure_block_response;
```
billmarriott-binary, line 132: Actually I was thinking of a flag `block_response` that would block all responses, success or failure. What do you think?

chylli-binary: Sorry I misunderstand your last comment. To be honest it is out of this card. I don't know if it will cause some strage behavior . But I think it is a good idea, and I guess it has no big problem.

chylli-binary: Let me do it now.

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -92,6 +92,8 @@ sub call_rpc {
     my $after_got_rpc_response_hook  = delete($req_storage->{after_got_rpc_response})  || [];
     my $before_call_hook             = delete($req_storage->{before_call})             || [];
     my $rpc_failure_cb               = delete($req_storage->{rpc_failure_cb});
+    # This flag will decide whether it should send back response.
```
michaelmueller-binary, line 95: can you expand this comment a bit.  Maybe add when this would be useful. and explain that it stops the response ever being sent to the client.

chylli-binary: Done.

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -92,6 +92,10 @@ sub call_rpc {
     my $after_got_rpc_response_hook  = delete($req_storage->{after_got_rpc_response})  || [];
     my $before_call_hook             = delete($req_storage->{before_call})             || [];
     my $rpc_failure_cb               = delete($req_storage->{rpc_failure_cb});
+    # If this flag true, then proxy will not send the rpc response to the client back.
```
michaelmueller-binary, line 95:
```suggestion
    # If this flag true, then proxy will not send the rpc response back to the clien.
```

lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm
```diff
@@ -92,6 +92,10 @@ sub call_rpc {
...
+    # It is very useful when websocket app itself (not websocket client) want to get information from rpc.
```
michaelmueller-binary, line 96:
```suggestion
    # It is very useful when the websocket app itself (not websocket client) wants to get information from rpc.
```

PR #69, opened 2020-04-27
Rpc queues/websocket proxy plugin

kirill-binary:


lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,351 @@
...
+keys in C<call_params>. 
+
+=item * C<rpc_response_callback>  If supplied this will be run with C<< Mojolicious::Controller >> instance the rpc_response and C<< $req_storage >>.
+B<Note:> if C<< rpc_response_callback >> is supplied the success and error callbacks are not used. 
+
+=item * C<before_get_rpc_response>  array ref of subroutines to run before the remote response, is passed C<< $c >> and C<< req_storage >>
+
+=item * C<after_get_rpc_response> arrayref of subroutines to run after the remote response,  is passed C<< $c >> and C<< req_storage >>
+called only when there is an actual response from the remote call .  IE if there is communication  error with the call it will
+not be called versus an error message being returned from the call when it will
```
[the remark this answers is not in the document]
kirill-binary: Agree, let me re-phrase it

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,350 @@
...
+=cut
+
+sub timeout {
+    return shift->{timeout} //= RESPONSE_TIMEOUT;
+}
+
+
+=head2 whoami
+
+Return uniq id of redis whick will be used by backend server to send repsonse.
```
turnstep, line 119:
```suggestion
Return unique ID of Redis which will be used by backend server to send response.
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,350 @@
...
+=item * C<call_params> a hashref of arguments on top of C<req_storage> to send to remote method. This will be suplemented with C<< $req_storage->{args} >>
+added as an C<args> key and be merged with C<< $req_storage->{stash_params} >> with stash_params overwriting any matching 
+keys in C<call_params>. 
+
+=item * C<rpc_response_callback>  If supplied this will be run with args: C<< Mojolicious::Controller >> instance, the rpc_response and C<< $req_storage >>.
+B<Note:> if C<< rpc_response_callback >> is supplied the success and error callbacks are not used. 
+
+=item * C<before_get_rpc_response>  array ref of subroutines to be run before the remote response is received, with args: C<< $c >> and C<< req_storage >>
+
+=item * C<after_get_rpc_response> arrayref of subroutines to be run after the remote response is received; called with args: C<< $c >> and C<< req_storage >>
```
turnstep, line 163: array ref or arrayref, for consistency. I've no strong feelings for which one :)

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,350 @@
...
+    my $rpc_response_cb = $self->get_rpc_response_cb($c, $req_storage);
+    my $before_get_rpc_response_hooks = delete($req_storage->{before_get_rpc_response}) || [];
+    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response})  || [];
+    my $before_call_hooks             = delete($req_storage->{before_call})             || [];
+    my $rpc_failure_cb                = delete($req_storage->{rpc_failure_cb});
+
+
+    foreach my $hook ($before_call_hooks->@*) { $hook->($c, $req_storage) }
+
+    $self->request($request_data)->then(sub {
```
turnstep, line 195: I'm guessing from this that perltidy hasn't got its hands on this file yet. :)

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,350 @@
...
+    return;
+}
+
+=head2 request
+
+Sends request to backend service. The method accepts single unnamed argument:
+
+=over 4
+
+=item * C<request_data> - should be C<arrayref>  wich should contain data for item which will be putted to redis stream.
```
turnstep, line 243: * putted

mat-fs:
```suggestion
=item * C<request_data> - an C<arrayref> containing data for the item which is going to be put into redis stream.
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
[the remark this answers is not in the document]
kirill-binary: Fixed

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,350 @@
...
+
+
+=head2 loop
+
+=cut
+
+sub loop {
+    my $self = shift;
+    return $self->{loop} //= do {
+        require IO::Async::Loop::Mojo;
```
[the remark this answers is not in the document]
kirill-binary: fixed, i removed `require`

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: Fixed, `arrayref`

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: Fixed

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,345 @@
...
+=back
+
+Returns undef.
+
+=cut
+
+sub call_rpc {
+    my ($self, $c, $req_storage) = @_;
+
+    # make sure that we are already waiting for messages
```
tom-binary, line 175: What does this refer to? There's no code next to these comments and it's not clear how it applies to the `->call_rpc` method itself?

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,345 @@
...
+
+    my ($msg_type, $request_data) = $self->_prepare_request_data($c, $req_storage);
+
+    my $rpc_response_cb = $self->get_rpc_response_cb($c, $req_storage);
+    my $before_get_rpc_response_hooks = delete($req_storage->{before_get_rpc_response}) || [];
+    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response})  || [];
+    my $before_call_hooks             = delete($req_storage->{before_call})             || [];
+    my $rpc_failure_cb                = delete($req_storage->{rpc_failure_cb});
+
+    foreach my $hook ($before_call_hooks->@*) { $hook->($c, $req_storage) }
```
tom-binary, line 186: odd - looks like there's something wrong with the perltidy rules, it should have flagged this or split onto multiple lines

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,345 @@
...
+
+    my $complete_future = $self->loop->new_future;
+
+    $self->wait_for_messages();
+
+    my $sent_future = $self->_send_request($request_data)->then(
+        sub {
+            my ($msg_id) = @_;
+
+            $self->pending_requests->{$msg_id} = $complete_future;
```
tom-binary, line 260: This looks like a race condition - I'd expect to be setting this *before* we send out the message.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,345 @@
...
+We'll use uniq id generated by L<whoami> as subscription channel.
+Subscription will be done only once within first request to backend server.
+
+=cut
+
+sub wait_for_messages {
+    my ($self) = @_;
+    $self->{already_waiting} //= $self->redis->subscribe(
+        [$self->whoami],
+        $self->$curry::weak(
```
tom-binary, line 299: should be able to write this as:

```
$self->redis->curry::weak::on(message => $self->curry::weak::_on_message)
```

I think?

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,345 @@
...
+
+    return;
+}
+
+sub _on_message {
+    my ($self, $redis, $raw_message) = @_;
+
+    my $message = eval { decode_json_utf8($raw_message) };
+
+    return unless ref $message eq 'HASH' && $message->{original_id};
```
tom-binary, line 313: This seems like an important case to be flagging, when might we expect non-hashref or missing `original_id` entries?

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,345 @@
...
+
+    my $completion_future = delete $self->pending_requests->{$message->{original_id}};
+
+    return unless $completion_future;
+
+    $completion_future->done($message);
```
tom-binary, line 319: I think this can fail if there was a timeout, since you can't `->done` a ready Future

t/26-rpc-queue-unit.t
```diff
@@ -0,0 +1,417 @@
...
+    my $rpc_id2 = $rpc_backend2->whoami;
+    ok $rpc_id2, 'Got rpc id';
+
+    is $rpc_id1, $rpc_backend1->whoami, 'Id will be unchanged for rpc instance';
+    is $rpc_id2, $rpc_backend2->whoami, 'Id will be unchanged for rpc instance';
+
+    isnt $rpc_id1, $rpc_id2, 'Id is uniq';
+};
+
+subtest _send_reuqest => sub {
```
tom-binary, line 71:
```suggestion
subtest _send_request => sub {
```

jy-binary:
please check tests

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: By specification we should use id, that we get from redis.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: in case of timeout we remove key with id from `$self->pending_requests`,  and this line `return unless $completion_future;` should protect us.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: probably you right, there is a room for race condition. Let me check with @ea-binary  about it.

ea-binary: @kirill-binary yes, let's change the specs.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: Fixed. i removed outdated comment

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: yep, you're right, this way it's looks much better.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: In normal situation we should never get this  kind of messages, but this is external data we have to validate it. Let me add logging for this case.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: Perl tidy is done, no changes for this line.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -0,0 +1,347 @@
...
+            my $api_response;
+
+            return Future->done unless $c && $c->tx;
+
+            if ($error eq 'Timeout') {
+                $api_response = $c->wsp_error($msg_type, 'RequestTimeout', 'Request is timed out.');
+            } else {
+                $api_response = $c->wsp_error($msg_type, 'WrongResponse', 'Sorry, an error occurred while processing your request.');
+            }
+            $rpc_failure_cb->($c, undef, $req_storage) if $rpc_failure_cb;
```
michaelmueller-binary, line 224: not passing message

kirill-binary: I tried to be consistent with job async, second argument should be rpc response, but in case of time out, it'll be undef. https://github.com/binary-com/perl-Mojo-WebSocketProxy/blob/master/lib/Mojo/WebSocketProxy/Backend/JobAsync.pm#L205

PR #70, opened 2020-07-22
ws/add_consumer_groups_backend

mehdi-fs:


lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -309,7 +311,7 @@ sub _on_message {
 
     my $message = eval { decode_json_utf8($raw_message) };
 
-    unless(ref $message eq 'HASH' && $message->{message_id}) {
+    if(ref $message ne ref {} && !$message->{message_id}) {
```
tom-binary, line 314: > `ref $message ne ref {}`

Please don't ever do this, it's only going to cause confusion...

We should have a separate `try` block for the JSON decoding, and return with an error on failure, then have a *separate* check for `message_id`.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -338,9 +340,9 @@ sub _prepare_request_data {
         [
         rpc      => $method,
         args     => encode_json_utf8($params),
```
tom-binary, line 342: This is conceptually wrong, the *entire* set of parameters needs to be UTF8-encoded. Doing partial encoding like this tends to lead to problems later if someone adds another key.

mehdi-fs: It going to be added to the stream by Redis `XADD` command. Since it accepts the even scalar parameters, so we have to encode arrays and hashes only. Actually, encoding at the moment is not a good idea unless we going to pass all of the parameters only in a pair of params to `XADD`.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
mehdi-fs: Changed and fixed, thanks

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -307,10 +312,19 @@ sub wait_for_messages {
...
+    }catch {
+        my $err = $@;
 
-    unless(ref $message eq 'HASH' && $message->{message_id}) {
-        $log->errorf('Fail to proccess response: %s', $raw_message);
+        $log->errorf('An error occurred while decoding published response by Consumer (Transport::Redis):', $err);
+        return;
+    }
+
+    if(!$message->{message_id}) {
```
kirill-binary, line 326: this data is from out side you have no guarantee that it'll be a hash. in case if it's a arrayref, you'll get exception here.

t/26-rpc-queue-unit.t
```diff
@@ -404,7 +403,7 @@ subtest 'RPC call: handling error response from rpc server' => sub {
 
     $cg_backend->call_rpc($c, $req_storage);
     $cg_backend->_on_message(undef,
-        qq[{"message_id": "$req_id", "result": { "error": { "code": "TestError", "message_to_client": "Error message", "details":{"field":"test_field"}}}}]
+        qq[{"message_id": "$req_id", "response": {"result": { "error": { "code": "TestError", "message_to_client": "Error message", "details":{"field":"test_field"}}}}}]
```
kirill-binary, line 406: looks horrible. Should we rename response to result?

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
mehdi-fs: How might it be `arrayref` since we always encode `hashrefs` before publish in the consumer part?

t/26-rpc-queue-unit.t
mehdi-fs: IDTS, we expect `response` as a key receive from the consumer.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
mat-fs: @mehdi-fs  This repository and module is open-source and independent of any business logic or company repos like `binary-websocket-api`. So it doesn't matter how `we` are using it; it should be developed and reviewed with putting our awareness of `binary-websocket-api` code in brackets.

mehdi-fs: @mat-fs I didn't know that, it makes sense, I will fix it

mat-fs: Yes. All repos under `binary-com` github company name are open source and public access.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -307,10 +312,19 @@ sub wait_for_messages {
 sub _on_message {
     my ($self, $redis, $raw_message) = @_;
 
-    my $message = eval { decode_json_utf8($raw_message) };
+    my $message = {};
+
+    try{
```
turnstep, line 317: Needs tidy

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -307,10 +312,19 @@ sub wait_for_messages {
...
+        my $err = $@;
 
-    unless(ref $message eq 'HASH' && $message->{message_id}) {
-        $log->errorf('Fail to proccess response: %s', $raw_message);
+        $log->errorf('An error occurred while decoding published response from worker:', $err);
+        return;
+    }
+
+    if(ref $message ne 'HASH' || !$message->{message_id}) {
+        $log->errorf('Failed to proccess response: message_id not exists at <%s>', $raw_message);
```
turnstep, line 327:
```suggestion
        $log->errorf('Failed to process response: message_id does not exist at <%s>', $raw_message);
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -334,14 +348,16 @@ sub _prepare_request_data {
...
+    my $request_data = [
         rpc      => $method,
-        args     => encode_json_utf8($params),
-        stash    => encode_json_utf8($stash_params),
         who      => $self->whoami,
         deadline => time + RESPONSE_TIMEOUT,
-        ];
+
+        $params  ?  (args   => encode_json_utf8 $params)        : (),
+        $stash_params   ?  (stash  => encode_json_utf8 $stash_params)  : (),
```
turnstep, line 357:
```suggestion
        $params  ?  (args   => encode_json_utf8($params))        : (),
        $stash_params   ?  (stash  => encode_json_utf8($stash_params))  : (),
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -277,7 +282,7 @@ sub _send_request {
 
     my $f = $self->loop->new_future;
     $self->redis->_execute(
-        xadd => XADD => ('rpc_requests', '*', $request_data->@*),
+        xadd => XADD => ('general', '*', $request_data->@*),
```
tom-binary, line 285: This change wasn't mentioned on the Trello card; please update the description.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -307,10 +312,19 @@ sub wait_for_messages {
...
+        my $err = $@;
 
-    unless(ref $message eq 'HASH' && $message->{message_id}) {
-        $log->errorf('Fail to proccess response: %s', $raw_message);
+        $log->errorf('An error occurred while decoding published response from worker:', $err);
+        return;
+    }
+
+    if(ref $message ne 'HASH' || !$message->{message_id}) {
+        $log->errorf('Failed to process response: message_id does not exist at <%s>', $raw_message);
```
tom-binary, line 327:
```suggestion
        $log->errorf('Failed to process response: message_id does not exist, original message content was %s', $raw_message);
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
mehdi-fs: There is no rule to tidy up the code, but change several lines manually.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
mehdi-fs: updated

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -307,10 +312,19 @@ sub wait_for_messages {
...
+    my $message = {};
+
+    try {
+        $message = decode_json_utf8($raw_message);
+    } catch {
+        my $err = $@;
 
-    unless(ref $message eq 'HASH' && $message->{message_id}) {
-        $log->errorf('Fail to proccess response: %s', $raw_message);
+        $log->errorf('An error occurred while decoding published response from worker:', $err);
```
turnstep, line 322:
```suggestion
        $log->errorf('An error occurred while decoding published response from worker: %s', $err);
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
mehdi-fs, line 285: Adding MAXLEN

chylli-binary (approved):

PR #71, opened 2020-09-07
rpc/handle_unwanted_ws_logs

mehdi-fs:


lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -201,7 +201,9 @@ sub call_rpc {
             }
             catch {
                 my $error = $@;
-                $rpc_failure_cb->($c, $result, $req_storage) if $rpc_failure_cb;
+                $rpc_failure_cb->($c, $result, $req_storage, {
+                        type => 'WrongResponse',
+                    }) if $rpc_failure_cb;
```
kirill-binary, line 206: we should have replicate same behavior like we have on JSONRPC
https://github.com/binary-com/perl-Mojo-WebSocketProxy/blob/master/lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm#L126

mehdi-fs: The only difference between them is error object params since other params (message, code) are extra params because they are useful only in case of `CallError`s unless we would like to keep behavior the same due to consistency.

kirill-binary: @mehdi-fs better to pass `code` and `message` in sake of consistency.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -193,22 +193,27 @@ sub call_rpc {
...
-                $rpc_failure_cb->($c, $result, $req_storage) if $rpc_failure_cb;
-                $api_response = $c->wsp_error($msg_type, 'WrongResponse', 'Sorry, an error occurred while processing your request.');
-            };
+            my $result = MojoX::JSON::RPC::Client::ReturnObject->new(rpc_response => $message->{response});
 
-            return Future->done  if $block_response || !$api_response;
+            foreach my $hook ($after_got_rpc_response_hooks->@*) { $hook->($c, $req_storage, $result) }
 
+            my $api_response;
+            if ($result->is_error){
```
kirill-binary, line 202: This method doesn't work, we're using incorrect implementation of jsonrpc.

kirill-binary: but probably nice to have this if to be consistent with jsonrpc implementation...

mehdi-fs: Thanks, the task not finished already and is in progress. will consider your comments.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -184,31 +182,36 @@ sub call_rpc {
 
     foreach my $hook ($before_call_hooks->@*) { $hook->($c, $req_storage) }
 
+    my ($msg_type, $request_data) = $self->_prepare_request_data($c, $req_storage);
```
turnstep, line 185: Why is this moved into the foreach?

kirill-binary: it's not moved into foreach , but moved after the loop.. 
Based on JSONRPC implementation it's not quite correct  we create jsonrpc request before calling `before_call_hooks`, which could affect request parametrs.
probably better to be consistent in this ordering with current implementation.

https://github.com/binary-com/perl-Mojo-WebSocketProxy/blob/master/lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm#L100-L107

turnstep: Oops, sorry, didn't expect perltidy to throw everything into one line there. :)

PR #74, opened 2020-09-24
rpc/separate_category_workers

nima-fs:


lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -73,6 +76,27 @@ sub loop {
     };
 }
 
+sub category_config {
+    my $self = shift;
+    return $self->{category_config} //= do {
+        my %processed = ();
+        try {
+            my $raw_config = YAML::XS::LoadFile($self->category_map_config);
```
kirill-binary, line 84: I'd try to keep this logic inside `binary-websocket-api`. and pass this configuration when we create backend. IMHO reading configuration is out of scope for this package

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -52,6 +53,8 @@ Creates object instance of the class
 
 =item * C<timeout> - Request timeout, in seconds. If not set, uses the environment variable C<RPC_QUEUE_RESPONSE_TIMEOUT>, or defaults to 300
 
+=item * C<category_map_config> - Category mapper configuration file path.
```
kirill-binary, line 56: we already have similar field in action config https://github.com/regentmarkets/binary-websocket-api/blob/master/lib/Binary/WebSocketAPI/Actions.pm#L243 why do we need to move it to separate yaml file?

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -174,17 +184,18 @@ Returns undef.
...
 
-    my $rpc_response_cb = $self->get_rpc_response_cb($c, $req_storage);
+    my $rpc_response_cb               = $self->get_rpc_response_cb($c, $req_storage);
     my $before_get_rpc_response_hooks = delete($req_storage->{before_get_rpc_response}) || [];
-    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response})  || [];
-    my $before_call_hooks             = delete($req_storage->{before_call})             || [];
+    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response}) || [];
+    my $before_call_hooks             = delete($req_storage->{before_call}) || [];
     my $rpc_failure_cb                = delete($req_storage->{rpc_failure_cb});
+    my $req_group                     = $req_storage->{msg_group} || 'general';
```
mehdi-fs, line 192: Please keep names consistent with the specifications and context of this task.

```suggestion
    my $category_name                 = $req_storage->{msg_group} || 'general';
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -257,7 +274,7 @@ And it'll be marked as failed in case of request timeout or in case of error put
 =cut
 
 sub request {
-    my ($self, $request_data) = @_;
+    my ($self, $request_data, $req_group) = @_;
```
mehdi-fs, line 277: Same 

```suggestion
    my ($self, $request_data, $req_category) = @_;
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -285,17 +303,19 @@ sub _next_request_id {
     return ++$self->{request_seq_id};
 }
 
+sub _rpc_category_timeout {
+    my ($self, $msg_group) = @_;
```
mehdi-fs, line 307: same

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -360,19 +379,19 @@ sub _prepare_request_data {
...
     my $msg_type = $req_storage->{msg_type} ||= $req_storage->{method};
 
-    my $params = $self->make_call_params($c, $req_storage);
+    my $params       = $self->make_call_params($c, $req_storage);
     my $stash_params = $req_storage->{stash_params};
 
     my $request_data = [
         rpc      => $method,
         who      => $self->whoami,
         deadline => time + RESPONSE_TIMEOUT,
```
mehdi-fs, line 391: The configured timeouts per categories must be considered here as well

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -285,17 +303,19 @@ sub _next_request_id {
     return ++$self->{request_seq_id};
 }
 
+sub _rpc_category_timeout {
+    my ($self, $msg_group) = @_;
+
+    my $rpc_group_timeout = $self->category_timeout_config->{$msg_group} // $self->timeout;
+    return $rpc_group_timeout;
```
mehdi-fs, line 310:
```suggestion
    return $self->category_timeout_config->{$category_name} // $self->timeout;
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -270,9 +287,10 @@ sub request {
 
     push @$request_data, ('message_id' => $msg_id);
 
-    my $sent_future = $self->_send_request($request_data);
+    my $rpc_category_timeout = $self->_rpc_category_timeout($req_group);
```
mehdi-fs, line 290: you can also ignore `rpc` in the names since they are obvious without that

```suggestion
    my $category_timeout = $self->_rpc_category_timeout($req_group);
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -52,6 +53,8 @@ Creates object instance of the class
 
 =item * C<timeout> - Request timeout, in seconds. If not set, uses the environment variable C<RPC_QUEUE_RESPONSE_TIMEOUT>, or defaults to 300
 
+=item * C<category_timeout_config> - Category timeouts configurations.
```
turnstep, line 56: Maybe reword - does not add any new information

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -174,17 +184,20 @@ Returns undef.
...
 
-    my $rpc_response_cb = $self->get_rpc_response_cb($c, $req_storage);
+    my $rpc_response_cb               = $self->get_rpc_response_cb($c, $req_storage);
     my $before_get_rpc_response_hooks = delete($req_storage->{before_get_rpc_response}) || [];
-    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response})  || [];
-    my $before_call_hooks             = delete($req_storage->{before_call})             || [];
+    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response}) || [];
+    my $before_call_hooks             = delete($req_storage->{before_call}) || [];
     my $rpc_failure_cb                = delete($req_storage->{rpc_failure_cb});
+    my $category_name                 = $req_storage->{msg_group} || 'general';
```
turnstep, line 192: Why are we not deleting this one too? (or perhaps the question is: why are we deleting those keys?

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -248,6 +267,10 @@ Sends request to backend service. The method accepts single unnamed argument:
 
 =item * C<request_data> - an C<arrayref> containing data for the item which is going to be put into redis stream.
 
+=item * C<req_category> - this will be passed to C<_send_request> to specify which redis category this message belongs to.
```
turnstep, line 270: Would be nice if these var names matched up with line 200

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
mehdi-fs: yeah better to indicates the default values or also an example of the YAML file structure will be good

mehdi-fs:
It's unrelated to the scope of this task, but please move L199 one line above, before `_prepare_request_data` statement.
https://github.com/binary-com/perl-Mojo-WebSocketProxy/blob/e300ef328e6406b5cbf81b7d1677826dff1b75b7/lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm#L199

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary, line 192: Lets move `general` to constant, also lets pass default category to back-end object.

nima-fs: Those key/values will be used to construct the `request` we are sending to rpc, so probably those other keys are not required and should be deleted, but as I didn't know if `msg_group` is being used in layers below, I didn't take the risk of deleting it. Would be nice if @kirill-binary could suggest if I can delete `msg_group` key as well or not.

nima-fs:
> It's unrelated to the scope of this task, but please move L199 one line above, before `_prepare_request_data` statement.
> https://github.com/binary-com/perl-Mojo-WebSocketProxy/blob/e300ef328e6406b5cbf81b7d1677826dff1b75b7/lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm#L199

:+1:

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -7,6 +7,7 @@ use Log::Any qw($log);
 use Mojo::Redis2;
 use IO::Async::Loop::Mojo;
 use Data::UUID;
+use YAML::XS;
```
mehdi-fs, line 10:
```suggestion
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -52,6 +54,10 @@ Creates object instance of the class
 
 =item * C<timeout> - Request timeout, in seconds. If not set, uses the environment variable C<RPC_QUEUE_RESPONSE_TIMEOUT>, or defaults to 300
 
+=item * C<category_timeout_config> - Hash containing the timeout value for each rpc call category.
+
+{ mt5 => 120, general => 5 }
```
mehdi-fs, line 59: one indent for pod renderer detect this line as code snippets
```suggestion
    { mt5 => 120, general => 5 }
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -52,6 +54,10 @@ Creates object instance of the class
 
 =item * C<timeout> - Request timeout, in seconds. If not set, uses the environment variable C<RPC_QUEUE_RESPONSE_TIMEOUT>, or defaults to 300
 
+=item * C<category_timeout_config> - Hash containing the timeout value for each rpc call category.
```
mehdi-fs, line 57:
```suggestion
=item * C<category_timeout_config> - A hash containing the timeout value for each request category.
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -112,6 +118,14 @@ sub timeout {
     return shift->{timeout} //= RESPONSE_TIMEOUT;
 }
 
+=head2 category_timeout_config
+
```
mehdi-fs, line 122: please add a line about the subroutine and its default value and what it returns

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -285,17 +310,18 @@ sub _next_request_id {
...
 
+sub _rpc_category_timeout {
+    my ($self, $category_name) = @_;
+
+    return $self->category_timeout_config->{$category_name} // $self->timeout;
+}
+
 sub _send_request {
-    my ($self, $request_data) = @_;
+    my ($self, $request_data, $req_group) = @_;
```
mehdi-fs, line 320:
```suggestion
    my ($self, $request_data, $category_name) = @_;
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -285,17 +310,18 @@ sub _next_request_id {
...
 
     my $f = $self->loop->new_future;
     $self->redis->_execute(
-        xadd => XADD => (
-            'general',
-            ('MAXLEN', '~', '100000'),
-            '*',
-            $request_data->@*
-        ),
+        xadd => XADD => ($req_group, ('MAXLEN', '~', '100000'), '*', $request_data->@*),
```
mehdi-fs, line 324: accordingly.
```suggestion
        xadd => XADD => ($category_name, ('MAXLEN', '~', '100000'), '*', $request_data->@*),
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -7,7 +7,7 @@ use Log::Any qw($log);
 use Mojo::Redis2;
 use IO::Async::Loop::Mojo;
 use Data::UUID;
-use JSON::MaybeUTF8 qw(encode_json_utf8 decode_json_utf8);
+    use JSON::MaybeUTF8 qw(encode_json_utf8 decode_json_utf8);
```
kirill-binary, line 10: ?

nima-fs: typo

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm.tidyup
```diff
@@ -0,0 +1,404 @@
+package Mojo::WebSocketProxy::Backend::ConsumerGroups;
```
mehdi-fs, line 1: this is committed wrongly.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -337,11 +366,10 @@ sub _on_message {
     }
 
     my $missed_param;
-    if(
-        ref $message ne 'HASH' 
-        || !$message->{$missed_param = "message_id"} 
-        || !$message->{$missed_param = "response"}
-    ) {
+    if (   ref $message ne 'HASH'
+        || !$message->{$missed_param = "message_id"}
```
tom-binary, line 370: Assigning to the same variable multiple times like this within the same expression is best avoided, please don't ever do this.

Instead, you can get a list of *all* the missing variables like this:

`my (@missing) = grep { !exists $message->{$_} } qw(message_id response)`

and the `ref $message ne 'HASH'` check should be split out anyway (that error makes no sense when it's not a hash).

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
kirill-binary: I'm not sure the reason why we `delete` callbacks i tryed to be consistent with HTTP transport.
https://github.com/binary-com/perl-Mojo-WebSocketProxy/blob/master/lib/Mojo/WebSocketProxy/Backend/JSONRPC.pm#L91-L94

IMHO `msg_group` we can keep. In theory it could be useful in hooks.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -52,6 +54,10 @@ Creates object instance of the class
 
 =item * C<timeout> - Request timeout, in seconds. If not set, uses the environment variable C<RPC_QUEUE_RESPONSE_TIMEOUT>, or defaults to 300
```
turnstep, line 55: Let's make this match the actual default while we are here (300 != 30)

nima-fs:
```suggestion
=item * C<timeout> - Request timeout, in seconds. If not set, uses the environment variable C<RPC_QUEUE_RESPONSE_TIMEOUT>, or defaults to 30
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -356,23 +387,24 @@ sub _on_message {
 }
 
 sub _prepare_request_data {
-    my ($self, $c, $req_storage) = @_;
+    my ($self, $c, $req_storage, $req_timeout) = @_;
 
     $req_storage->{call_params} ||= {};
+    $req_timeout //= RESPONSE_TIMEOUT;
```
billmarriott-binary, line 393: This looks unnecessary, `$req_timeout` cannot be undef here.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
chylli-binary: Why there are 2 same pod section ?  https://github.com/binary-com/perl-Mojo-WebSocketProxy/pull/74/files#diff-9e7d49105b98b317451e638ee705edadfafb05e3a9925b625ec533a3f2fd7bafR57

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -174,17 +191,20 @@ Returns undef.
...
 
-    my $rpc_response_cb = $self->get_rpc_response_cb($c, $req_storage);
+    my $rpc_response_cb               = $self->get_rpc_response_cb($c, $req_storage);
     my $before_get_rpc_response_hooks = delete($req_storage->{before_get_rpc_response}) || [];
-    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response})  || [];
-    my $before_call_hooks             = delete($req_storage->{before_call})             || [];
+    my $after_got_rpc_response_hooks  = delete($req_storage->{after_got_rpc_response}) || [];
+    my $before_call_hooks             = delete($req_storage->{before_call}) || [];
     my $rpc_failure_cb                = delete($req_storage->{rpc_failure_cb});
+    my $category_name                 = $req_storage->{msg_group} || DEFAULT_CATEGORY_NAME;
```
chylli-binary, line 199: Is it possible to use same word of hash key `msg_group` and variable name `category_name` ? how about `msg_category` ?

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -248,6 +274,10 @@ Sends request to backend service. The method accepts single unnamed argument:
```
michaelmueller-binary, line 274: Line 271 needs to be updated for the new arguments.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -21,6 +21,8 @@ no indirect;
 __PACKAGE__->register_type('consumer_groups');
 
 use constant RESPONSE_TIMEOUT => $ENV{RPC_QUEUE_RESPONSE_TIMEOUT} // 30;
+use constant DEFAULT_CATEGORY_NAME => 'general';
```
michaelmueller-binary, line 24: Seems that this should be documented somewhere that it will default to `general`?

nima-fs: We are just switching `mt5` calls in this stage and here: https://github.com/regentmarkets/binary-websocket-api/blob/master/lib/Binary/WebSocketAPI/Actions.pm#L89, we only set `msg_group` for mt5 methods. So to keep the change non-breaking and not requiring to set `general` for all other methods in `Actions.pm`, I thought it was pretty much obvious that the default value should be `general`. If this should get documented somewhere, please guide me on where to add this.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -336,13 +367,14 @@ sub _on_message {
...
-        || !$message->{$missed_param = "response"}
-    ) {
-        $log->errorf("Failed to process response: '%s' does not exist, original message content was %s", $missed_param, $raw_message);
+    if (ref $message ne 'HASH') {
+        $log->errorf("Failed to process response: Invalid message type got: %s, want: HASH", ref $message);
+        return;
+    }
+
+    my (@missing_params) = grep { !exists $message->{$_} } REQUIRED_RESPONSE_PARAMETERS;
+    if (scalar @missing_params) {
```
zakame, line 376: No need for explicit `scalar`, without any other expressions the if conditional already implies scalar/boolean context:
```suggestion
    if (@missing_params) {
```

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -285,17 +315,18 @@ sub _next_request_id {
...
 
     my $f = $self->loop->new_future;
     $self->redis->_execute(
-        xadd => XADD => (
-            'general',
-            ('MAXLEN', '~', '100000'),
-            '*',
-            $request_data->@*
-        ),
+        xadd => XADD => ($category_name, ('MAXLEN', '~', '100000'), '*', $request_data->@*),
```
zakame, line 329: This nested list `('MAXLEN', '~', '100000')` doesn't really do what you'd expect; Perl will flatten it into the outer list (along with the contents of `$request_data->@*`.)  We could however use `qw(MAXLEN ~ 100000)` so as to reduce the need for quotes.

```suggestion
        xadd => XADD => ($category_name, qw(MAXLEN ~ 100000), '*', $request_data->@*),
```

Also, since our minimum Perl for this public CPAN module is `5.014`, we probably should bump https://github.com/binary-com/perl-Mojo-WebSocketProxy/blob/71fbbcf23c7dd5d1097dd8b9ed55b40f1e168ed0/cpanfile#L14 to `5.024` use `postderef`, otherwise we can't use it; seems we've had other postderef usage in other parts here but our dependencies have not been updated to reflect this change.

mehdi-fs (approved):

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
nima-fs: Addressed.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
chylli-binary: But I don't think it is a bad habit.

zakame:
Will need corresponding `Changes` updates.

lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
michaelmueller-binary: Is this going to be a public module?

zakame: This _is_ a public module: https://metacpan.org/pod/Mojo::WebSocketProxy

Its probably best to have for `ConsumerGroups->new` to accept a `msg_group` or `category` parameter that can fall back to `general`, so it is better documented here (and less associated with a specific category implementation.)

nima-fs: I think it is currently a public module :thinking:

nima-fs: @zakame I don't think we want to run category-specific websocket proxies! This fallback here is for the actions that don't have their `msg_group` set. So I don't see how passing a `msg_group` to `ConsumerGroups` constructor can help improve the situation.

michaelmueller-binary: In any case the public will have no idea how we are implementing it so it needs needs to be documented that it will use `general` by default.

zakame: @nima-fs my point is exactly what @michaelmueller-binary said: if its a publicly visible change, it needs to be documented.

(Also, need to refrain from discussing further specifics or linking to private repos here, again this is a public module!)

nima-fs: I have deleted the comment that included link to private repos. Will add the documentation as well. @zakame @michaelmueller-binary

PR #87, opened 2022-04-21
myriad support

mahrous-deriv:


lib/Mojo/WebSocketProxy/Backend/ConsumerGroups.pm
```diff
@@ -364,7 +365,9 @@ Subscription will be done only once within first request to backend server.
 
 sub wait_for_messages {
     my ($self) = @_;
-    $self->{already_waiting} //= $self->redis->subscribe([$self->whoami], $self->redis->curry::weak::on(message => $self->curry::weak::_on_message));
+    $self->{already_waiting} //=
+        $self->redis->subscribe(['myriad.' . $self->whoami, $self->whoami],
```
billmarriott-binary, line 369: Is it possible to only add the prefix based on `msg_group` so we subscribe to one channel?

raunakkathuria, line 369: Let's not add `myriad` hardcoded in proxy, it's an open source module. Pass it as a config or as a parameter

mahrous-deriv: We can pass it from websocket.conf.

mahrous-deriv: Yes, we can also pass the prefix and prepend to the `$self->whoami` in `_prepare_request_data`

PR #95, opened 2023-10-16
log/request_logger_implementation

aisha-deriv:
This is the extended version of reference [POC implementation](https://github.com/hani-deriv/perl-Mojo-WebSocketProxy/pull/1) to associate a wrapper for logging into request storage we can use the context to be appended in the message with Deriv adapter and in this PR we will set context at time of dispatch of request and remove once the message is logged for the request

clickup card : https://app.clickup.com/t/20696747/PLA-660


lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -123,6 +124,9 @@ sub on_message {
     my $req_storage = {};
     $req_storage->{args} = $args;
 
+    # any further key-val pairs can be added here in request storage context object
+    $req_storage->{logger_context} = {correlation_id => UUID::Tiny::create_UUID_as_string(UUID::Tiny::UUID_V4)};
```
hani-deriv, line 128: we probably can move the context creation to the logger itself. if we found that there is no performance hit by keeping this constructor here.

lib/Mojo/WebSocketProxy/RequestLogger.pm
```diff
@@ -0,0 +1,30 @@
...
+
+# Method to log messages with various log levels
+method log_message($level, $message, @params) {
+    if($log->adapter->can('set_context')) {
+        $log->adapter->set_context($context);
+    }
+    $log->$level($message, @params);
+    if($log->adapter->can('clear_context')) {
+        $log->adapter->clear_context;
+    }
```
chylli-deriv, line 23: why we need set_context and clear_context ? 
why not 
```
local $log->context->{$_} = $context->{$_} for keys %$context;
```

aisha-deriv: we are using the Deriv adapter function here for setting the context in log object and then clearing after its usage

chylli-deriv: I don't think that's a good  idea to clear_context. That will make log mess. Suppose you want to print a context all time:
```
$log->context->{hello} = "world";
```
and after that you want to set context:
```
$log->adapter->set_context({foo => "bar"});
$log->info("blahblah");
$log->adapter->clear_context;
```
then after that the context 'hello' wlll be discarded.
that's not a good idea.
And, I guess we shouldn't call $log->adapter in most cases.

what we should do is:

```
for my $key (keys %$context){
    local $log->context->{$key} = $context->{$key};
}
$log->$level($message, @params);

```

hani-deriv: There are two contexts @chylli-deriv:
`$log->context`: is the one the module itself supports. it's global and limited as per the current implementation of the module and doing changes on it is a bit tricky.
`$log->adapter->context` is the one introduced by @aisha-deriv and we have a bit of control over it only in the `DERIV` adapter without affecting the complete module.
We can probably also provide support for the global context inside the new one.

hani-deriv: @aisha-deriv we can introduce an accessor method to the adapter and expose the context:
```
sub context {my $self = shift; return $self->{context}};
```
Then we can happily take advantage of the `local` keyword.

Another way around the local is to define a setter and clearer for the RequestLogger in which it sets and clears its own context out of the adapter context.

lib/Mojo/WebSocketProxy/RequestLogger.pm
```diff
@@ -0,0 +1,70 @@
...
+}
+
+method warn($message) {
+    $self->log_message('warning', $message);
+}
+
+method debug($message) {
+    $self->log_message('debug', $message);
+}
+
```
hani-deriv, line 65: We have `crit` and `critf` also.

lib/Mojo/WebSocketProxy/RequestLogger.pm
aisha-deriv: @chylli-deriv  as @hani said we are using adapter context because of the format we have applied to it and its not global
@hani-deriv adding this accessor will not lead issues with random context values ? with the clearing context we are always sure that correct context is put in message

lib/Mojo/WebSocketProxy/RequestLogger.pm
aisha-deriv: ok I'll add this

lib/Mojo/WebSocketProxy/RequestLogger.pm
hani-deriv: We extend the functionality only, and we keep the ability to completely clear the context.
Depending on the usage could be random or *global* values.
For the `RequestLogger` if it sets its context and clears it. no chance for these random values, unless something went wrong.

lib/Mojo/WebSocketProxy/Dispatcher.pm
```diff
@@ -123,6 +123,7 @@ sub on_message {
     my $req_storage = {};
     $req_storage->{args} = $args;
 
+    $req_storage->{logger}         = Mojo::WebSocketProxy::RequestLogger->new(context => {});
```
hani-deriv, line 126: Not used. 
```suggestion
    $req_storage->{logger}         = Mojo::WebSocketProxy::RequestLogger->new;
```

lib/Mojo/WebSocketProxy/RequestLogger.pm
```diff
@@ -0,0 +1,70 @@
...
+# Method to log messages with various log levels
+method log_message($level, $message, @params) {
+    if($log->adapter->can('set_context')) {
+        $log->adapter->set_context($context);
+    }
+    $log->$level($message, @params);
+    if($log->adapter->can('clear_context')) {
+        $log->adapter->clear_context;
+    }
```
kirill-deriv, line 23: @aisha-deriv it's a very strange choice for can, you're testing statically named method.. and without check executing dynamic method... it should be other way around.

lib/Mojo/WebSocketProxy/RequestLogger.pm
aisha-deriv: There is always a room for enhancement we can cater the request for localize the context as @chylli-deriv mentioned later to the adapter. For now we strictly want to keep the context availability similar for all messages so the scope is defined to be per request for log messages.

lib/Mojo/WebSocketProxy/RequestLogger.pm
aisha-deriv: this is done @kirill-deriv , added a default handler that will be used in case the context adapter not available this is done with regards to the discussion of @hani-deriv POC mentioned in PR description

lib/Mojo/WebSocketProxy/RequestLogger.pm
```diff
@@ -0,0 +1,91 @@
...
+    my ($level, $message, $context, @params) = @_;
+
+    if(scalar @params) {
+        return $log->$level($message, @params);
+    }
+
+    return $log->$level($message, $context);
+};
+
+sub set_handler {
```
leonid-deriv, line 22: Can we initialise it during object construction. Do we really need to change the handler?

t/29-request_logger_test.t
```diff
@@ -0,0 +1,27 @@
...
+    $logger->warn('This is a warning message');
+    $logger->trace('This is a trace message');
+    $logger->error('This is an error message');
+    $logger->debug('This is debug message');
+    $logger->infof('This is a info message %s', ['abc', '221']);
+    $logger->warnf('%s This is a warning message ', {with => 'params'});
+    $logger->tracef('This is a trace %s message', "with params");
+    $logger->errorf('This is an error %s message',"with params");
+    $logger->debugf('This is debug message with %s', 'params');
+    pass('All log levels tested');
```
leonid-deriv, line 19: how do we test it. just make sure that we implemented these methods?

aisha-deriv: we are checking the implementation for warn message in another test https://github.com/binary-com/perl-Log-Any-Adapter-DERIV/pull/20/files#diff-5ce3bf7ac545da06738a7184a78470998388c0ec9e97849cc9b67b0dce1a826a, all other are being put here to check methods are implemented

lib/Mojo/WebSocketProxy/RequestLogger.pm
hani-deriv: The `RequestLogger` resides in `WebsocketProxy` which is a public module. We don't want to tie it up to our special `DERIV` logging handling logic.
The object will be created on each message and we don't really need to change the handler. setting it once on start-up of the worker is enough.

lib/Mojo/WebSocketProxy/RequestLogger.pm
```diff
@@ -0,0 +1,91 @@
...
+class Mojo::WebSocketProxy::RequestLogger;
+
+use strict;
+use warnings;
+use Log::Any qw($log);
+use UUID::Tiny;
+
+field $context;
+
+our $default_handler = sub {
```
hani-deriv, line 12: IMO, When we call it, it'll be slightly misleading. maybe `log_handler` is a better name.

aisha-deriv: agreed! will change it

t/29-request_logger_test.t
```diff
@@ -0,0 +1,27 @@
+use Test::More;
+use Mojo::WebSocketProxy::RequestLogger;
+
+my $context = {'correlation_id' => '1234'};
+my $logger = Mojo::WebSocketProxy::RequestLogger->new(context => $context);
```
hani-deriv, line 5: The constructor doesn't accept the context as param.

chylli-deriv (approved):

kirill-deriv (approved):

leonid-deriv (approved):

